PR #3, opened 2021-09-08
Measure availability group

lmd59:
# Summary
Creates the measure availability test group in the deqm test kit. Uses a constant of available measures, which should in the future be pulled from the embedded cqf_ruler instance to populate a drop down (to be added as an inferno core UI element in the future).
## New behavior
deqm test kit provides includes test group "Measure Availability" which may be selected and tested with a cqf ruler instance as the client. Preloaded cqf_ruler should pass. A server without EXM130 would fail.
## Code changes
Adds `measure_availability.rb` test group and `measure_availability_spec.rb` to test pass/fail expectations. Updates the main deqm test kit file to use the new measure availability test group.
# Testing guidance
This can be tested by running the test kit (`docker-compose build`, `docker-compose pull`, `docker-compose up`), and using a cqf ruler instance (`docker run -d -p 3002:8080 tacoma/cqf-ruler-preloaded:0.5.0.no-vs`) or cqf_ruler embedded (as long as it starts properly on `docker-compose up`). The cqf-ruler url looks something like `http://{ip}:{port}/cqf-ruler-r4/fhir` and this is used as the url input when running the test through the inferno UI, which is accessed at `http://{ip}:4567`.

### DIFF
--- a/lib/deqm_test_kit/measure_availability.rb
+++ b/lib/deqm_test_kit/measure_availability.rb
@@ -30,10 +30,7 @@ module DEQMTestKit
         assert_resource_type(:bundle)
         assert_valid_json(response[:body])
         measure_bundle = JSON.parse(response[:body])
-        
-
         assert measure_bundle["total"].positive?, "Expected to find measure with identifier #{measure_identifier} and version #{measure_version}"
-        assert false, "Measure bundle logged #{measure_bundle.to_s}"
       end
     end
   end
--- a/lib/deqm_test_kit/measure_availability.rb
+++ b/lib/deqm_test_kit/measure_availability.rb
@@ -1,0 +1,40 @@
+module DEQMTestKit
+  class MeasureAvailability < Inferno::TestGroup
+    id 'measure_availability'
+    title 'Measure Availability'
+    description 'Ensure selected measures are available on the fhir server'
+
+    fhir_client do
+      url :url
+    end
+
+    MEASURES = ['EXM130|7.3.000','EXM125|7.3.000']
+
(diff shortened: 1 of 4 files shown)


lib/deqm_test_kit/measure_availability.rb
```diff
@@ -0,0 +1,39 @@
...
+  class MeasureAvailability < Inferno::TestGroup
+    id 'measure_availability'
+    title 'Measure Availability'
+    description 'Ensure selected measures are available on the fhir server'
+
+    fhir_client do
+      url :url
+    end
+
+    MEASURES = ['EXM130|7.3.000', 'EXM125|7.3.000'].freeze
```
natjoe4, line 14: It doesn't look like the EXM125 module is ever used. Does it make sense to keep it in for future testing purposes? I guess the way we pull in measures will change soon (hopefully) regardless.

natjoe4 (approved):
Functionality looks good to me. I tested it pretty rigorously throughout my task and again for the review. I left one comment as a possible clean-up that doesn't affect the functionality.

sarahmcdougall (approved):
This is my first time looking at deqm-test-kit, but the syntax all looks good and logical to me. Functionality is good as well. Great work!

lib/deqm_test_kit/measure_availability.rb
lmd59: Yeah- this is more of a temporary example to reflect how a set of measures will be used to populate the dropdown and then one of them will be selected by the user

PR #4, opened 2021-09-09
Data req suite

natjoe4:
# Summary
deqm-test-kit can now test data-requirements operations of a test client server against the data-requirements output of the embedded cqf-ruler instance
## New behavior
There is a new test category: "Data Requirements" in the DEQM Measure Operations Test Suite when running the deqm-test-kit using docker=compose commands which will run a data-requirements operation on a hard-coded (for now) measure on both the passed in test client and the embedded cqf-ruler instance. The test will pass if the results are retrieved from both sources and match. 
## Code changes
- Added a new class DataRequirements
- Added code to deqm-test-kit.rb to include DataRequirements checks in the test suite
- Added unit tests for DataRequirements test module
# Testing guidance
1. Start a Docker Daemon running in the background (I did this by opening Docker desktop)
2. run `docker-compose build`
3. run `docker-compose pull`
4. run `docker-compose up`
5. navigate to [http://localhost:4567](http://localhost:4567)
6. select DEQM Measure Operations Test Suite and click "Start Testing"
7. Wait for cqf_ruler_to finish start-up. You should see `Server:main: Started @ {timestamp}` logged to the terminal
8. Click the play button by Data Requirements
9. After a moment, this test should pass and display a green checkmark

In addition to testing functionality, please examine the code in deqm_test_kit/data_requirements.rb closely. I have not worked with Ruby or Inferno before and had limited testing options, so any input here would be greatly appreciated. A couple of notes before diving in:

1. The fhir_search function is repeating code from the measure_availability test module. The ultimate goal here is to force the tests to be run in sequence and save the measure id retrieved in the measure_availability test in a place where the data_requirements test can access it. For the time being, we don't know how to do this. This may become a subsequent task
2. The Measure to test is currently hard coded. Eventually we will want the user to be able to select which measure to use for the data_requirements operation via a dropdown in the front-end. I believe we are waiting for the Inferno team to build this functionality
3. The embedded cqf_ruler URL is currently hard-coded as well as a string. Ideally we want to replace this with some symbol that stores the URL. I am unsure how to do this for now, but happy to work through it with a reviewer/make changes if someone knows how to accomplish this.

### DIFF
--- a/lib/deqm_test_kit/data_requirements.rb
+++ b/lib/deqm_test_kit/data_requirements.rb
@@ -1,73 +1,81 @@
+# frozen_string_literal: true
+
 module DEQMTestKit
+  # GET [base]/Measure/CMS146/$data-requirements?periodStart=2014&periodEnd=2014
+  class DataRequirements < Inferno::TestGroup
+    id 'data_requirements'
+    title 'Data Requirements'
+    description 'Ensure fhir server can respond to the $data-requirements request'
 
-    #GET [base]/Measure/CMS146/$data-requirements?periodStart=2014&periodEnd=2014
-    class DataRequirements < Inferno::TestGroup
-      id 'data_requirements'
-      title 'Data Requirements'
-      description 'Ensure fhir server can respond to the $data-requirements request'
-  
-      fhir_client do
-        url :url
-      end
+    fhir_client do
+      url :url
+    end
 
-      fhir_client :embedded_client do
-        url 'http://cqf_ruler:8080/cqf-ruler-r4/fhir'
-      end
-  
-      test do
-        title 'Check data requirements against expected return'
-        id 'data-requirements-01'
-        description 'Data requirements on the fhir test server match the data requirements of our embedded client'
-        makes_request :data_requirements
+    fhir_client :embedded_client do
+      url 'http://cqf_ruler:8080/cqf-ruler-r4/fhir'
+    end
 
-        PARAMS = {
-            resourceType: "Parameters",
-            parameter:[{
-                periodStart: '2019-01-01',
-                periodEnd: '2019-12-31'
-              
-            }]
-      }
-        run do
-          # Look for matching measure from cqf-ruler datastore by resource id
-          # TODO: actually pull measure from user input drop down (populated from embedded client)
-          measure_to_test = 'EXM130|7.3.000'
-          measure_identifier, measure_version = measure_to_test.split('|')
-    
-          # Search system for measure by identifier and version
-          fhir_search(:measure, params: { name: measure_identifier, version: measure_version }, name: :measure_search)
-          assert_response_status(200)
-          assert_resource_type(:bundle)
-          assert_valid_json(response[:body])
-          measure_bundle = JSON.parse(response[:body])
-          assert measure_bundle["total"].positive?, "Expected to find measure with identifier #{measure_identifier} and version #{measure_version}"
-          
-          id = measure_bundle["entry"][0]["resource"]["id"]
+    PARAMS = {
+      resourceType: 'Parameters',
+      parameter: [{
+        periodStart: '2019-01-01',
+        periodEnd: '2019-12-31'
 
-          # Run our data requirements operation on the test client server
-          fhir_operation("Measure/#{id}/$data-requirements", body: PARAMS, name: :data_requirements)
-          library = JSON.parse(response[:body])
-          actual_dr = library["dataRequirement"]
-          assert_response_status(200)
-          assert_resource_type(:library)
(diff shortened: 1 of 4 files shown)


spec/deqm_test_kit/data_requirements_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+      test_measure_response = FHIR::Bundle.new(total: 1, entry: [{ resource: { id: test_id } }])
+      test_library_response = FHIR::Library.new(dataRequirement: [{ type: 'Condition' }])
+
+      stub_request(:get, "#{url}/Measure?name=#{measure_name}&version=#{measure_version}")
+        .to_return(status: 200, body: test_measure_response.to_json)
+
+      stub_request(:get, "#{embedded_client}/Measure?name=#{measure_name}&version=#{measure_version}")
+        .to_return(status: 200, body: test_measure_response.to_json)
+
+      stub_request(:post, "#{url}/Measure/#{test_id}/$data-requirements")
```
natjoe4, line 37: Can someone explain why the data-requirements operation ends up being a POST request? I thought data-requirements was supposed to be a GET, but the outgoing request from fhir_operation is always a POST

lib/deqm_test_kit/data_requirements.rb
```diff
@@ -0,0 +1,81 @@
...
+
+        # Run data requirements operation on embedded cqf-ruler instance
+        fhir_operation("Measure/#{id}/$data-requirements", body: PARAMS, client: :embedded_client,
+                                                           name: :data_requirements)
+        expected_library = JSON.parse(response[:body])
+        expected_dr = expected_library['dataRequirement']
+
+        # Ensure both data requirements results libraries are identical
+        assert((expected_dr - actual_dr).blank?,
+               "Client data-requirements is missing expected data requirements for measure #{id}")
```
lmd59, line 74: I know this is the same as the previous data_requirement implementation, but the id here is actually the embedded server's measure id rather than the test client's measure id. I would potentially name those ids two different things and use the test client id here since they can actually differ.

spec/deqm_test_kit/data_requirements_spec.rb
lmd59: The inferno-core operation implementation just default [does a post](https://github.com/inferno-community/inferno-core/blob/c4081927733881e3624c24eb78959f7ca06daa65/lib/inferno/dsl/fhir_client.rb#L80). Inferno team did say it was a bit immature. You're right that this operation would typically be a GET (https://www.hl7.org/fhir/measure-operation-data-requirements.html). I wonder if this has the potential to trip up non-cqf_ruler servers.

spec/deqm_test_kit/data_requirements_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    let(:test) { group.tests.first }
+    let(:measure_name) { 'EXM130' }
+    let(:measure_version) { '7.3.000' }
+    let(:test_id) { 'test_id' }
+
+    it 'passes if a Library was received' do
+      test_measure_response = FHIR::Bundle.new(total: 1, entry: [{ resource: { id: test_id } }])
+      test_library_response = FHIR::Library.new(dataRequirement: [{ type: 'Condition' }])
+
+      stub_request(:get, "#{url}/Measure?name=#{measure_name}&version=#{measure_version}")
```
lmd59, line 31: Since some of these stubs remain the same for every `it` check, can we pull them out under the `describe` umbrella so that they can be used across each check's context?

spec/deqm_test_kit/data_requirements_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    Inferno::Repositories::TestRuns.new.results_for_test_run(test_run.id)
+  end
+
+  describe 'data requirements test' do
+    let(:test) { group.tests.first }
+    let(:measure_name) { 'EXM130' }
+    let(:measure_version) { '7.3.000' }
+    let(:test_id) { 'test_id' }
+
+    it 'passes if a Library was received' do
```
lmd59, line 27:
```suggestion
    it 'passes if the expected Library was received' do
```

lmd59: This is a nitpick

lmd59 (approved):

spec/deqm_test_kit/data_requirements_spec.rb
lmd59: This has proven a little bit tricky to address atm, so we can look into sharing `stub_request` s later

mgramigna (approved):

PR #6, opened 2021-09-20
Submit data

lmd59:
# Summary
Changes from the connectathon and updates for cleaner usage of inferno core. Adds deqm-test-server and mongo to docker compose.
## New behavior
Adds submit data test group and passes data through from previous test groups to support submit data.
Starts deqm-test-server as a part of the docker compose.
Disables patient group and patient group tests.
## Code changes

- Updates docker-compose.
- Updates groups in deqm_test_kit file.
- Expands the capabilities of measure_availability and data_requirements to support submit data and utilizes more inferno core built ins.
- Adds submit_data test group, which leverages basic measure information and queries built from the data requirements to pull resources from the embedded cqf_ruler and submit them to the client under test using the `$submit-data` operation. Then checks that submitted resources can be found by searching the client under test.
- Updates and adds tests accordingly.

# Testing guidance
All tests can be run with the `rspec` command. To run the application, use `docker-compose up --build`, which may take a minute or more to start up (cqf-ruler will show a time taken until start). Access the application at http://localhost:4567 . Select the test suite and start testing. Starting with the capability statement or measure availability test, input the test url and test with either:

- http://deqm_test_server:3000/4_0_0 (needs to be loaded with at least measure 130 from connectathon via POST to http://localhost:3000/4_0_0, if you have problems loading the measure due to size, ask channel for workaround, there may be a new test server image to use )
- http://cqf_ruler:8080/cqf-ruler-r4/fhir (if cqf_ruler, there may be a bug in the preloaded image, so if you'd like to use this, you can change the image to `contentgroup/cqf-ruler:latest` in the `docker-compose.yml` and post the 130 measure bundle from connectathon to load it with data)

Once measure availability is run, it should pass the correct information through to run the data requirements test group. Data requirements should then pass through necessary information to submit data. Should look something like this:
Screen Shot 2021-09-28 at 10 08 33 PM
Screen Shot 2021-09-28 at 10 08 52 PM

The data requirements test should succeed with cqf_ruler, but fail with deqm_test_server because there are still some differences between how these servers create data requirements. All other inferno tests should pass. Feel free to play with other patient data, but this currently only works with 130 (until we get a measure dropdown later).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -47,3 +47,23 @@ services:
     command: bundle exec sidekiq -r ./worker.rb
     depends_on:
       - redis
+  deqm_test_server:
+    depends_on:
+      - mongo
+    image: tacoma/deqm-test-server:connectathon
+    environment:
+      DB_HOST: mongo
+      DB_PORT: 27017
+      DB_NAME: deqm-test-server-dev
+    ports:
+      - "3000:3000"
+    command: npm start
+  mongo:
+    image: mongo:4.4.4
+    ports:
+      - "27017:27017"
+    volumes:
+      - mongo_data:/data/db
+
+volumes:
+  mongo_data:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,7 +50,7 @@ services:
   deqm_test_server:
     depends_on:
       - mongo
-    image: tacoma/deqm-test-server:connectathon
+    image: tacoma/deqm-test-server:latest
     environment:
       DB_HOST: mongo
       DB_PORT: 27017
--- a/lib/deqm_test_kit.rb
+++ b/lib/deqm_test_kit.rb
@@ -3,6 +3,7 @@
 require_relative 'deqm_test_kit/patient_group'
 require_relative 'deqm_test_kit/measure_availability'
 require_relative 'deqm_test_kit/data_requirements'
+require_relative 'deqm_test_kit/submit_data'
 
 module DEQMTestKit
   class Suite < Inferno::TestSuite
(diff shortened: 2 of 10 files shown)


sarahmcdougall (approved):
All the tests passed for me and I was able to run the application and have it work as expected. I am not too familiar with Ruby, so I’m not sure if you want a second reviewer to look more closely at the code, but it looked good to me! Thank you for including the images of the expected test inputs.

docker-compose.yml
```diff
@@ -47,3 +47,23 @@ services:
     command: bundle exec sidekiq -r ./worker.rb
     depends_on:
       - redis
+  deqm_test_server:
+    depends_on:
+      - mongo
+    image: tacoma/deqm-test-server:connectathon
```
mgramigna, line 53:
```suggestion
    image: tacoma/deqm-test-server:latest
```
suggest using the `latest` tag that is auto-built on deqm-test-server docker hub now

mgramigna (changes requested):
Everything looks good. Added one small comment

Also, note, I got request entity too large for the $submit-data call to deqm-test-server. This is because our middleware for increasing the request entity size only handles `application/json+fhir`, not `application/fhir+json` lol. So this is a problem with the test server itself, that we can resolve separately. 

Everything else looks good here

docker-compose.yml
lmd59: ah smart- forgot about that one

mgramigna (approved):
yayyyhyyyyyyyyy

PR #10, opened 2021-11-01
Measure availability negative test

natjoe4:
# Summary
Added a new integration test to the measure availability set which expects a measure-availability request to return an empty bundle when called with a measure id that does not have a corresponding measure in the system

## New behavior
Inferno measure-availability suite should not cover the test case where the serve does not have the requested measure available

## Code changes
Added a new test case to the measure-availability suite

# Testing guidance
- Start up the test kit using `docker-compose up --build`
- Navigate to [localhost:4567](localhost:4567)
- select 'DEQM Measure Operations Test Suite' and click 'start testing'
- click the play button next to Measure Availability and ensure that the popup in the bottom left corner shows 2 tests to be run.
- Check that both pass!

### DIFF
--- a/lib/deqm_test_kit/measure_availability.rb
+++ b/lib/deqm_test_kit/measure_availability.rb
@@ -41,7 +41,7 @@ module DEQMTestKit
       title 'Measure cannot be found returns empty bundle'
       id 'measure-availability-02'
       description 'Selected measure is know not to exist on the server and returns an empty bundle'
-      uses_request :measure_search
+      makes_request :measure_search_failure
       output :null
 
       run do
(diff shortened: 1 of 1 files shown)


lib/deqm_test_kit/measure_availability.rb
```diff
@@ -36,5 +36,27 @@ class MeasureAvailability < Inferno::TestGroup
         output measure_id: resource.entry[0].resource.id
       end
     end
+
+    test do
+      title 'Measure cannot be found returns empty bundle'
+      id 'measure-availability-02'
+      description 'Selected measure is know not to exist on the server and returns an empty bundle'
+      uses_request :measure_search
```
mgramigna, line 44: I don't actually think we want the `uses_request` pattern here. This will be useful when we want to do something with the response returned for this named request from within another test group without having to duplicate the request. However, this test makes a different request and does not need to rely on the response from the `measure_search` in the previous test.

This actually causes an issue where this test can't be run standalone since it depends on the `measure_search` request having been made by the previous test, even though that isn't actually a dependency of this test

mgramigna (changes requested):

mgramigna (approved):

PR #11, opened 2021-11-02
Data requirements negative test cases

natjoe4:
# Summary
Added negative test cases to data requirements suite

## New behavior
Inferno integration tests will now check that the server correctly returns a 400 when a data requirements request is submitted without the appropriate parameters or with an invalid measure id

## Code changes
Added tests to data-requirements suite. Updated previous test to correctly pass in query parameters

# Testing guidance
-run `rspec` and ensure all tests pass
- run `docker-compose pull`
- run `docker-compose up --build`
- navigate to [localhost:4567](http://localhost:4567) after startup of cqf-ruler
- select DEQM Measure Operations Test Suite and click start
- click the play button next to Measure Availability and input `http://deqm_test_server:3000/4_0_1`
- Once the test has passed, click the play button next to Data Requirements. Ensure that it runs three tests. The first one should fail (since our output varies compared to cqf-ruler), but the other two should pass.
- Now, return to [localhost:4567](http://localhost:4567), select DEQM Measure Operations Test Suite and click start
- Repeat the above but replace our test server url with `http://cqf_ruler:8080/cqf-ruler-r4/fhir`
- Now the data requirements first test should pass, but the next two should fail. This is because cqf-ruler returns the wrong error codes.

### DIFF
--- a/lib/deqm_test_kit/data_requirements.rb
+++ b/lib/deqm_test_kit/data_requirements.rb
@@ -21,17 +21,7 @@ module DEQMTestKit
 
     PARAMS = {
       resourceType: 'Parameters',
-      parameter: [{
-        periodStart: '2019-01-01',
-        periodEnd: '2019-12-31'
-      }]
-    }.freeze
-
-    MISSING_PARAMS = {
-      resourceType: 'Parameters',
-      parameter: [{
-        periodStart: '2019-01-01'
-      }]
+      parameter: [{}]
     }.freeze
 
     INVALID_ID = 'INVALID_ID'
@@ -96,6 +86,7 @@ module DEQMTestKit
                "Client data-requirements contains unexpected data requirements for measure #{measure_id}: #{diff}")
       end
     end
+
     test do
       title 'Check data requirements returns 400 for missing parameters'
       id 'data-requirements-02'
@@ -103,13 +94,14 @@ module DEQMTestKit
 
       run do
         # Run our data requirements operation on the test client server
-        fhir_operation("Measure/#{INVALID_ID}/$data-requirements", body: MISSING_PARAMS)
+        fhir_operation("Measure/#{INVALID_ID}/$data-requirements", body: PARAMS)
         assert_response_status(400)
         assert_valid_json(response[:body])
         assert(resource.resourceType == 'OperationOutcome')
         assert(resource.issue[0].severity == 'error')
       end
     end
+
     test do
       title 'Check data requirements returns 400 for invalid measure id'
       id 'data-requirements-03'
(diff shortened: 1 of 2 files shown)


lib/deqm_test_kit/data_requirements.rb
```diff
@@ -83,6 +86,39 @@ class DataRequirements < Inferno::TestGroup
...
     end
+
+    test do
+      title 'Check data requirements returns 400 for missing parameters'
+      id 'data-requirements-02'
+      description 'Data requirements returns 400 when periodStart and periodEnd parameters are omitted'
+
+      run do
+        # Run our data requirements operation on the test client server
+        fhir_operation('Measure/TEST_ID/$data-requirements', body: PARAMS)
```
sarahmcdougall, line 97: Instead of `TEST_ID`, I think it would be better to use `measure_id` so that there is only one issue with this test case (missing params) and not two with the addition of a technically invalid id

spec/deqm_test_kit/data_requirements_spec.rb
```diff
@@ -160,4 +191,94 @@ def run(runnable, inputs = {})
...
+  describe '$data-requirements with missing parameters' do
+    let(:test) { group.tests[1] }
+    let(:measure_name) { 'EXM130' }
+    let(:measure_version) { '7.3.000' }
+    let(:measure_id) { 'measure-EXM130-7.3.000' }
+
+    it 'passes with correct OperationOutcome returned' do
+      stub_request(
+        :post,
+        "#{url}/Measure/TEST_ID/$data-requirements"
```
sarahmcdougall, line 204: Same comment as above with replacing the `TEST_ID` here

spec/deqm_test_kit/data_requirements_spec.rb
```diff
@@ -160,4 +191,94 @@ def run(runnable, inputs = {})
...
+      )
+        .to_return(status: 200, body: error_outcome.to_json)
+      result = run(test, url: url, measure_id: measure_id)
+      expect(result.result).to eq('fail')
+    end
+    it 'fails when resource returned is not of type OperationOutcome' do
+      stub_request(
+        :post,
+        "#{url}/Measure/TEST_ID/$data-requirements"
```
sarahmcdougall, line 222: Same comment as above

spec/deqm_test_kit/data_requirements_spec.rb
```diff
@@ -160,4 +191,94 @@ def run(runnable, inputs = {})
...
+      )
+        .to_return(status: 400, body: test_library_response.to_json)
+      result = run(test, url: url, measure_id: measure_id)
+      expect(result.result).to eq('fail')
+    end
+    it 'fails when severity of OperationOutcome returned is not of type error' do
+      stub_request(
+        :post,
+        "#{url}/Measure/TEST_ID/$data-requirements"
```
sarahmcdougall, line 231: Same comment as above

sarahmcdougall (changes requested):
Everything looks good to me and runs properly! I only have one suggestion that we talked about over Slack with making sure the id is valid for the negative param tests. Good work!

sarahmcdougall (approved):

PR #12, opened 2021-11-03
Submit data negative test cases

natjoe4:
# Summary
Added test cases which expect the server to return errors for known bad $submit-data requests

## New behavior
When running integration tests, the $submit-data suite will now execute 3 tests instead of 1.

## Code changes

- Added test to check for 400 returned for Parameters object submitted with no MeasureReport
- Added test to check 400 returned for Parameters object submitted with multiple MeasureReports
- Added rspec tests to check correct behavior of above tests

# Testing guidance
run `rspec` to ensure all unit tests pass
Follow the steps outline [here](https://github.com/projecttacoma/deqm-test-kit/pull/11).
Then, click the play button next to submit data and click submit
Ensure the suite runs 3 tests
- Submit Data valid submission
- fails if a measureReport is not submitted
- fails if multiple measureReports are submitted

All tests should pass!

### DIFF
--- a/lib/deqm_test_kit/submit_data.rb
+++ b/lib/deqm_test_kit/submit_data.rb
@@ -111,6 +111,7 @@ module DEQMTestKit
         end
       end
     end
+
     test do
       title 'fails if a measureReport is not submitted'
       id 'submit-data-02'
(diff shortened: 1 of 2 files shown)


mgramigna (approved):
LGTM, tests behave as expected against test server

PR #15, opened 2021-11-17
Bulk import

MegRiley:
# Summary
Added a test group for the bulk import support
## New behavior
In the test group menu there is now the option to run a test for bulk import support
## Code changes
Added two new files : 
bulk_import.rb - this file contains the bulk of the changes to support a bulk data test  on the deqm test server
bulk_import_utils.rb - this file contains the back off loop logic (It was largely grabbed from the old deqm-test-client work done for the ATL connectathon in 2019)

Modified the deqm-test-kit.rb file to now include the bulk_import group of tests
# Testing guidance
Pull, and build. 
Launched inferno, it is import to execute the measure availability test first. (the bulk_import test is similar to submit data in that it won't work without that test having been run)
Once the measure availablity test has been run against deqm-test-server, select bulk import.
When the menu comes up the measure should be populated and then you just need to set the server to deqm-test-server and the test should run and complete without issue.  (Note the  suite is currently failing because the server is returning a 500, this is not a problem with the tests but with the smart of fhir server)


lib/deqm_test_kit/bulk_import.rb
```diff
@@ -0,0 +1,69 @@
...
+require_relative '../utils/bulk_import_utils'
+module DEQMTestKit
+  # BulkImport test group ensure the fhir server can accept bulk data import requests
+  class BulkImport < Inferno::TestGroup
+    id 'bulk_import'
+    title 'Bulk Import'
+    description 'Ensure the fhir server can accept bulk data import requests'
+
+    input :measure_id
+    custom_headers = { 'X-Provenance': '{"resourceType": "Provenance"}', prefer: 'respond-async' }
```
lmd59, line 12: Is there a reason one of the keys is a string and the other is a symbol?

lib/deqm_test_kit/bulk_import.rb
```diff
@@ -0,0 +1,69 @@
...
+    params = {
```
lmd59, line 13: Could potentially use FHIR::Parameters.new() to build this for a little bit of easier maintenance and property verification in code.

lib/deqm_test_kit/bulk_import.rb
```diff
@@ -0,0 +1,69 @@
...
+    # rubocop:disable Metrics:BlockLength
+    test do
+      title 'Ensure data can be accepted'
+      id 'bulk-import-01'
+      description 'Send the data to the server and the response is a 202 (or is it?)'
+      run do
+        assert(measure_id,
+               'No measure selected. Run Measure Availability prior to running the bulk import test group.')
+        fhir_read(:measure, measure_id)
+        assert_valid_json(response[:body])
```
lmd59, line 42: Are we checking that the measure exists on the server here? If so, I would probably assert a 200 response.

lib/deqm_test_kit/bulk_import.rb
```diff
@@ -0,0 +1,69 @@
...
+        fhir_operation("Measure/#{measure_id}/$submit-data", body: params, name: :submit_data)
+        reply = fhir_client(:url).send(:get, '$bulkstatus')
```
lmd59, line 44: do we need to pass in url here since our default fhir_client already uses `:url` (line 30)? Also, just to confirm, I'm guessing we're doing this as a send instead of a fhir_operation because the test server does not support an empty body $bulkstatus as a post?

lib/deqm_test_kit/bulk_import.rb
```diff
@@ -0,0 +1,69 @@
...
+        polling_url = url + reply.headers('Content-Location')
+        fhir_operation('$bulkstatus', polling_url)
```
lmd59, line 46: Do we want to do anything with the response to this request?

lib/deqm_test_kit/bulk_import.rb
```diff
@@ -0,0 +1,69 @@
...
+        loop do
+          reply = nil
+          begin
+            reply = fhir_client.client.get(polling_url)
+          rescue RestClient::TooManyRequests => e
+            reply = e.response
+          end
+          wait_time = get_retry_or_backoff_time(wait_time, reply)
+          seconds_used = Time.now - start
+          # exit loop if we get a successful response or timeout reached
```
lmd59, line 60: Is this saying that a successful response is anything that isn't a 202 or 429? What about a 500 or 404, etc?

lmd59:
Also recommend adding some spec tests in spec/deqm_test_kit

lmd59:
I did get an issue when testing:
"Error: no block given

/opt/inferno/vendor/bundle/ruby/2.7.0/gems/inferno_core-0.0.7/lib/inferno/dsl/fhir_client_builder.rb:12:in `instance_exec'"
Does this look familiar?

lib/deqm_test_kit/bulk_import.rb
MegRiley: The only reason was that rubocop was complaining about the 2nd header.

lib/deqm_test_kit/bulk_import.rb
MegRiley: Yes, I can add the assert 200, I had just used the same logic the submit data suite was using.

lib/deqm_test_kit/bulk_import.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+require_relative '../utils/bulk_import_utils'
+require 'byebug'
```
natjoe4, line 4:
```suggestion

```

lib/deqm_test_kit/bulk_import.rb
natjoe4: Looks like this is still relevant. Unless there was an issue with this approach?

lib/deqm_test_kit/bulk_import.rb
```diff
@@ -0,0 +1,73 @@
...
+      run do
+        assert(measure_id,
+               'No measure selected. Run Measure Availability prior to running the bulk import test group.')
+        fhir_read(:measure, measure_id)
+        assert_valid_json(response[:body])
+        fhir_operation("Measure/#{measure_id}/$submit-data", body: params, name: :submit_data)
+        location_header = response[:headers].find { |h| h.name == 'content-location' }
+        # temporary fix for extra 4_0_1
+        polling_url = "#{url}/#{location_header.value.sub('4_0_1/', '')}"
+        reply = get(polling_url)
```
natjoe4, line 49: It doesn't look like this reply variable is ever used. I think it can be removed entirely

lib/utils/bulk_import_utils.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module DEQMTestKit
+  # Utility functions in support of the bulk import test group
+  module BulkImportUtils
+    def get_retry_or_backoff_time(wait_time, reply)
+      retry_after = -1
+      is_retry_nil = reply[:headers].find { |h| h.name == 'retry_after' }
```
natjoe4, line 8:
```suggestion
      is_retry_nil = reply[:headers].find { |h| h.name == 'Retry-After' }
```

natjoe4, line 8: The header here is supposed to be `Retry-After`, but not sure what the implications of changing this are for the functionality of the code. I would try changing it locally to make sure it doesn't break everything.

spec/deqm_test_kit/bulk_import_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    Inferno::TestRunner.new(test_session: test_session, test_run: test_run).run(runnable)
+  end
+
+  describe 'The server is able to perform bulk data tasks' do
+    let(:test) { group.tests[0] }
+    let(:measure_name) { 'EXM130' }
+    let(:measure_version) { '7.3.000' }
+    let(:measure_id) { 'measure-EXM130-7.3.000' }
+    url = 'http://example.com/fhir'
+    it 'can proceed since the measure exists' do
```
natjoe4, line 25:
```suggestion
    it 'passes on successful $bulkImport' do
```

spec/deqm_test_kit/bulk_import_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      resource = FHIR::Bundle.new(total: 1, entry: [{ resource: { id: 'test_id' } }])
+
+      stub_request(:get, "#{url}/Measure/#{measure_id}")
+        .to_return(status: 200, body: test_measure.to_json)
+
+      stub_request(:post, "#{url}/Measure/#{measure_id}/$submit-data")
+        .to_return(status: 200, body: resource.to_json, headers: { 'content-location': 'location' })
+      polling_url = "#{url}/location"
+      stub_request(:get, polling_url)
+        .to_return(status: 202, body: resource.to_json).times(3)
```
natjoe4, line 36: If you do want to retry based on the `Retry-After` header, you could mock this call to return a specific `Retry-After` header

natjoe4 (changes requested):
Functionality looks good! Left a few small comments ad it seems like some of Lauren's comments are still relevant

lib/deqm_test_kit/bulk_import.rb
MegRiley: What this is saying is that loop will break if its not a 202 or a 429, the thinking is that if the server is sending a 500 or a 404 the import has been disrupted and isn't likely complete successfully so the test will fail.

natjoe4 (changes requested):
Sorry should have caught these in the last review, but super small changes and then this should be good!

lib/deqm_test_kit/bulk_import.rb
```diff
@@ -0,0 +1,71 @@
...
+        polling_url = "#{url}/#{location_header.value.sub('4_0_1/', '')}"
+        get(polling_url)
+        wait_time = 1
+        start = Time.now
+        seconds_used = 0
+        timeout = 120
+        loop do
+          begin
+            get(polling_url)
+          rescue RestClient::TooManyRequests => e
```
natjoe4, line 56: I believe that TooManyRequests errors will return a Reply-After header anyway that can get processed by the existing get_retry_or_backoff_time code, so this rescue is unnecessary.

lib/deqm_test_kit/bulk_import.rb
```diff
@@ -0,0 +1,71 @@
...
+      run do
+        assert(measure_id,
+               'No measure selected. Run Measure Availability prior to running the bulk import test group.')
+        fhir_read(:measure, measure_id)
+        assert_valid_json(response[:body])
+        fhir_operation("Measure/#{measure_id}/$submit-data", body: params, name: :submit_data)
+        location_header = response[:headers].find { |h| h.name == 'content-location' }
+        # temporary fix for extra 4_0_1
+        polling_url = "#{url}/#{location_header.value.sub('4_0_1/', '')}"
+        get(polling_url)
```
natjoe4, line 48:
```suggestion
        
```
We call get(polling_url) twice here before processing the response.

lib/deqm_test_kit/bulk_import.rb
MegRiley: Yeah when I pushed the change to this, I broke the spec tests even after following the example in the submit data tests, I'm not sure if it was the way I was doing it or it was just the way the data was formatted

natjoe4 (approved):
Looks good to me!

PR #16, opened 2021-11-17
Patient/$everything test

sarahmcdougall:
# Summary
This PR creates the Patient/$everything test group in the deqm test kit. 

## New behavior
The Patient/$everything test group has been created. It tests valid `Patient/<id>/$everything` and `Patient/$everything` requests, as well as an invalid request to a non-existent patient ID in the database.

## Code changes
The `deqm_test_kit/patient_everything.rb` file creates the `PatientEverything` test group and three tests for the cases outlines above. The `patient_everything_spec` creates RSpec tests for both positive and negative test cases. 

# Testing guidance
All tests can be run using the `rspec` command. To run the application, make sure you have docker running, then run the commands `docker-compose pull` followed by `docker-compose up --build`. Navigate to `http://localhost:4567` and select the test suite. It may take a few moments for cqf ruler to get started up. Click the play button for Patient/$everything and check that the tests pass. For the url, use `http://deqm_test_server:3000/4_0_1`.


lib/deqm_test_kit/patient_everything.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    TEST_PATIENT_ID = 'test-patient'
+
+    test do
+      title 'Patient/<id>/$everything valid submission'
+      id 'patient-everything-01'
+      description 'Patient data is received for single patient on the server'
+      makes_request :patient_everything
+      output :queries_json
```
natjoe4, line 23: you can get rid of the output here. This test isn't returning queries_json

lib/deqm_test_kit/patient_everything.rb
```diff
@@ -0,0 +1,91 @@
...
+        assert(resource.entry.count { |x| x.resource.resourceType == 'Patient' } == 1)
+      end
+    end
+
+    test do
+      title 'Patient/$everything valid submission'
+      id 'patient-everything-02'
+      description 'Patient data is received for all patients on the server'
+      makes_request :patient_everything
+      output :queries_json
```
natjoe4, line 51: you can get rid of the output here. This test isn't returning queries_json

lib/deqm_test_kit/patient_everything.rb
```diff
@@ -0,0 +1,91 @@
...
+      url :url
+    end
+
+    TEST_PATIENT_ID = 'test-patient'
+
+    test do
+      title 'Patient/<id>/$everything valid submission'
+      id 'patient-everything-01'
+      description 'Patient data is received for single patient on the server'
+      makes_request :patient_everything
```
natjoe4, line 22: The use of makes_request in this suite isn't quite correct. makes_request is used to save requests that you may want to reuse later. Since I don't think we are reusing any of the requests in this suite you can get rid of all the makes_requests and the names in the fhir operations for each of these tests

lib/deqm_test_kit.rb
```diff
@@ -45,5 +46,6 @@ class Suite < Inferno::TestSuite
     group from: :measure_availability
     group from: :data_requirements
     group from: :submit_data
+    group from: :patient_everything
```
natjoe4, line 49: Nit picking here, but I think it would be best to have patient_everything be the first test in the group (after Capability Statement). Since measure_availability -> data_req -> submit_data all need to be run in sequence and the url for patient_everything gets populated after running the capability statement suite anyway. This is just my opinion though, so your call. If you do choose to do this the suite number will need to be changed in all the rspec tests

natjoe4 (changes requested):
This is great work. Left a few inline comments that range from inconsequential to severe nitpicks. Functionality all works as described.

lib/deqm_test_kit.rb
sarahmcdougall: You know this better than I do so I'll trust your judgment

natjoe4 (approved):
No notes here. Nice job!

PR #17, opened 2021-11-30
Evaluate Measure Negative Tests

sarahmcdougall:
# Summary
This PR creates negative tests and RSpec (unit) tests for the `$evaluate-measure` operation.

## New behavior
The `$evaluate-measure` and `$care-gaps` operations now have basic tests thanks to the great Matt Gramigna.

The `$evaluate-measure` test group contains positive and negative tests. For positive tests, the test group checks that individual, population, and subject-list reports can be created (NOTE: the subject-list test will fail for deqm-test-server). There is also a positive test for the support of non-required params (NOTE: this will fail in deqm-test-server because some of the non-required params are not currently supported). The negative requests cover various situations: missing required params, missing `subject` param for an `individual` `reportType`, and the inclusion of an invalid `reportType`.

## Code changes
The `deqm_test_kit/evaluate_measure.rb` file contains the positive and negative tests for the operation. The `evaluate_measure_spec.rb` file creates RSpec tests for both positive and negative test cases.

# Testing guidance
All tests can be run using the `rspec` command. To run the application, make sure you have `docker` running, then run the commands `docker-compose pull` followed by `docker-compose up --build`. Navigate to `http://localhost:4567` and select the test suite. Click the play button for `$evaluate-measure` and check that the tests pass. For the url, use `http://deqm_test_server:3000/4_0_1`. I used `measure-EXM130-7.3.000` for the measure id and `numer-EXM130` for the patient id.

### DIFF
--- a/lib/deqm_test_kit/care_gaps.rb
+++ b/lib/deqm_test_kit/care_gaps.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 module DEQMTestKit
+  # tests for $care-gaps
   class CareGaps < Inferno::TestGroup
     id 'care_gaps'
     title 'Gaps in Care'
@@ -21,7 +22,8 @@ module DEQMTestKit
       input :period_end, default: '2019-12-31'
 
       run do
-        fhir_operation("/Measure/$care-gaps?measureId=#{measure_id}&periodStart=#{period_start}&periodEnd=#{period_end}&subject=#{patient_id}&status=open")
+        params = "measureId=#{measure_id}&periodStart=#{period_start}&periodEnd=#{period_end}&subject=#{patient_id}"
+        fhir_operation("/Measure/$care-gaps?#{params}&status=open")
 
         assert_response_status(200)
         assert_resource_type(:bundle)
--- a/lib/deqm_test_kit/care_gaps.rb
+++ b/lib/deqm_test_kit/care_gaps.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module DEQMTestKit
+  class CareGaps < Inferno::TestGroup
+    id 'care_gaps'
+    title 'Gaps in Care'
+    description 'Ensure FHIR server can calculate gaps in care for a measure'
+
+    fhir_client do
+      url :url
+    end
+
+    INVALID_ID = 'INVALID_ID'
+
+    test do
+      title 'Check $care-gaps proper calculation'
+      id 'care-gaps-01'
+      description 'Server should properly return a gaps report'
+      input :measure_id, :patient_id
+      input :period_start, default: '2019-01-01'
+      input :period_end, default: '2019-12-31'
+
+      run do
+        fhir_operation("/Measure/$care-gaps?measureId=#{measure_id}&periodStart=#{period_start}&periodEnd=#{period_end}&subject=#{patient_id}&status=open")
+
+        assert_response_status(200)
+        assert_resource_type(:bundle)
+        assert_valid_json(response[:body])
+      end
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


lib/deqm_test_kit/care_gaps.rb
```diff
@@ -0,0 +1,34 @@
...
+      url :url
+    end
+
+    INVALID_ID = 'INVALID_ID'
+
+    test do
+      title 'Check $care-gaps proper calculation'
+      id 'care-gaps-01'
+      description 'Server should properly return a gaps report'
+      input :measure_id, :patient_id
```
lmd59, line 20: This is probably out of scope for your task, but for future tasking, should we have something like a patient availability test for the patient id (similar to measure availability)?

lib/deqm_test_kit/evaluate_measure.rb
```diff
@@ -0,0 +1,168 @@
...
+      input :period_start, default: '2019-01-01'
+      input :period_end, default: '2019-12-31'
+
+      run do
+        params = "periodStart=#{period_start}&periodEnd=#{period_end}&subject=#{patient_id}"
+        fhir_operation("/Measure/#{measure_id}/$evaluate-measure?#{params}")
+
+        assert_response_status(200)
+        assert_resource_type(:measure_report)
+        assert_valid_json(response[:body])
```
lmd59, line 31: Again, not sure this is in your task scope, but do we want to verify the accuracy of the evaluate-measure results in some way? Do we want to compare against cqf-ruler or anything like that?

lib/deqm_test_kit/care_gaps.rb
```diff
@@ -0,0 +1,34 @@
...
+      input :period_start, default: '2019-01-01'
+      input :period_end, default: '2019-12-31'
+
+      run do
+        params = "measureId=#{measure_id}&periodStart=#{period_start}&periodEnd=#{period_end}&subject=#{patient_id}"
+        fhir_operation("/Measure/$care-gaps?#{params}&status=open")
+
+        assert_response_status(200)
+        assert_resource_type(:bundle)
+        assert_valid_json(response[:body])
```
lmd59, line 30: Again, out of scope for your task, but $care-gaps test fails for a numerator patient (works for a denominator patient). I know that numerator patient wouldn't be expected to have gaps, but doesn't seem like the test should fail for that case.

lmd59 (approved):
I got nothin' for your code. Magnifique. All of my comments are for potential expansions to the positive test cases, which I don't think are your problem, but thought I'd put in here so we can think about if they're worthy of future tasking.

lib/deqm_test_kit/evaluate_measure.rb
```diff
@@ -32,9 +33,48 @@ class EvaluateMeasure < Inferno::TestGroup
...
+      input :period_start, default: '2019-01-01'
+      input :period_end, default: '2019-12-31'
+
+      run do
+        params = "periodStart=#{period_start}&periodEnd=#{period_end}&subject=#{patient_id}&reportType=subject-list"
+        fhir_operation("/Measure/#{measure_id}/$evaluate-measure?#{params}")
+
+        assert_response_status(200)
+        assert_resource_type(:measure_report)
+        assert_valid_json(response[:body])
```
lmd59, line 51: should we just quick check that the report type is subject list here as well?

lib/deqm_test_kit/evaluate_measure.rb
```diff
@@ -32,9 +33,48 @@ class EvaluateMeasure < Inferno::TestGroup
...
+      input :period_start, default: '2019-01-01'
+      input :period_end, default: '2019-12-31'
+
+      run do
+        params = "periodStart=#{period_start}&periodEnd=#{period_end}&reportType=population"
+        fhir_operation("/Measure/#{measure_id}/$evaluate-measure?#{params}")
+
+        assert_response_status(200)
+        assert_resource_type(:measure_report)
+        assert_valid_json(response[:body])
```
lmd59, line 69: check report type is population type

lmd59 (approved):

PR #18, opened 2021-12-07
Care gaps negative

natjoe4:
# Summary
Added tests to $care-gaps suite to check our responses in situations when the $care-gaps operation is expected to fail

## New behavior
The $care-gaps suite now contains six tests instead of one, 5 of which check for the correct response after an invalid $care-gaps request

## Code changes
Added new tests to `care_gaps.rb` that check for the correct response after invalid care-gaps requests
Added rspec testing to test the new tests

# Testing guidance
- Run the unit tests with `rspec`
- Make sure docker is running, then boot up the test kit using `docker-compose up --build`
- Run the `Measure Availability` suite with `http://deqm_test_server:3000/4_0_1` as the url
- Run the `Gaps in Care` suite using the following parameters
- measure_id: measure-EXM130-7.3.000
- patient_id: numer-EXM130
- period_start: 2019-01-01
- period_end: 2019-12-31
- We should expect to see 3 failures here, 2 because out test server doesn't yet support certain properties it should, and 1 because a numerator patient on a $care-gaps operation will return an empty object, not a bundle. This trips the assertion in the suite which expects a bundle returned


MegRiley (approved):
Looks good to me

PR #20, opened 2022-01-03
Regular $import test group

sarahmcdougall:
# Summary
Added test group for $import and renamed the existing bulk import test group to "bulk_submit_data."

## New behavior
Test group functionality was added for $import. Once the test kit is up and running, you will see the new test group, titled "Non-Measure-Specific Bulk Import." When a url is provided (http://deqm_test_server:3000/4_0_1), a POST request will be sent to `$import`.

## Code changes
The new `bulk_import.rb` file contains the test group for non-specific bulk import. This test group is very similar to the "bulk submit data" test group, except it
- does not require a measure
- calls $import in the `fhir_operation` instead of `Measure/id/$submit-data`
- does not take in a `measureReport` object in the `parameter` array

# Testing guidance
Run the `rspec` tests and ensure that they all pass. Start up the test kit with `docker-compose up --build` and check that the non-measure-specific bulk import test group appears. When running the test, note that a 500 error will be thrown - check that this is the same 500 error that comes up for the bulk submit data test group (error about Redis connection).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,11 +50,14 @@ services:
   deqm_test_server:
     depends_on:
       - mongo
+      - redis
     image: tacoma/deqm-test-server:latest
     environment:
       DB_HOST: mongo
       DB_PORT: 27017
       DB_NAME: deqm-test-server-dev
+      REDIS_HOST: redis
+      REDIS_PORT: 6379
     ports:
       - "3000:3000"
     command: npm start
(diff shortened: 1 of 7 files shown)


natjoe4 (changes requested):
These changes look good to me. I was able to get the tests to pass after manipulating the URL to add _type=Patient, but it was failing before then. Is it possible to add a user input bar to take in a string of types and append them to the end of the URL? I think this is the intended endgame with $import anyway. The smart-health-it server takes too long to export all types causing the timeout to expire for me.

natjoe4 (approved):
Looks great to me. Love the type parameter addition!

PR #21, opened 2022-01-19
Changes from connectathon

mgramigna:
# Summary

Not too many updates here, just addressing some of the TODO workarounds that we initially had and updating the request bodies to be proper.

## New behavior

* Bulk submit data should fail with 500 after a few polls because `_typeFilter` is invalid
* Regular bulk import should pass entirely

## Code changes

* Update environment of deqm-test-server to reflect new environment variables
* `valueString` -> `valueUrl` where applicable
* Update parsing of `content-location` header to use as a full URL since it is no longer a slug when returned from the server
* Update tests with proper content-location header

# Testing guidance

Test the two bulk data groups:

1. Test the bulk submit-data with measure id `measure-EXM130-7.3.000` (ensure you have posted this measure bundle to deqm-test-server beforehand). Should get 500 with error message that the export server gave us 404 (this is because it doesn't support `_typeFilter` which is default behavior in bulk submit data!
2. Test the non measure-specific bulk $import group with whatever types you want, e.g. `Patient`; `Patient,Encounter,Procedure`, completely blank. It will take a little while but should pass and get an operation ndjson URL in the response.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,12 +53,16 @@ services:
       - redis
     image: tacoma/deqm-test-server:latest
     environment:
+      HOST: deqm_test_server
+      PORT: 3000
       DB_HOST: mongo
       DB_PORT: 27017
       DB_NAME: deqm-test-server-dev
       REDIS_HOST: redis
       REDIS_PORT: 6379
       NODE_TLS_REJECT_UNAUTHORIZED: 0
+      IMPORT_WORKERS: 2
+      NDJSON_WORKERS: 2
     ports:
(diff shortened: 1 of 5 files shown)


lmd59 (approved):

lmd59 (changes requested):
Issue was discovered with not clearing out old type parameters. Not specifically an issue with deqm-test-kit, but we might want to hold off on merging this until addressing the issue or making a new task to address the issue

lmd59 (approved):

PR #22, opened 2022-01-26
Regular import bug fix

sarahmcdougall:
# Summary
Changes were made to the regular bulk import test so that the `valueUrl` properly resets with each new test.

## New behavior
Previously, the functionality was fine when running the test once. However, when running the test a second time, the `valueUrl` on the parameters object would not reset. For example, if we ran the test once with type = "Patient," and then we ran the test again with type = "Procedure," the test would end up with a `valueUrl` like this: `[url]/$export/_type=Patient_type=Procedure`. This would result in a 500 error being thrown.

Now, the `params` object successfully resets at the end of the test, and so we can run the test multiple times without getting any errors.

## Code changes
I moved the `params` object so that it is within the test, rather than just being within the class. 

I made some changes to how the `valueUrl` is actually updated. Turns out, the `types` input gets set to an empty string if the user does not specify any types in the UI. Therefore, I changed the if statement to check that the length of `types` is positive before changing the `valueUrl`. I think empty string is truthy in Ruby, so the previous check `if types` was landing us in the if block, and so `"type="` was being tacked onto the `valueUrl`. 

Also, the use of `concat` was causing the `valueUrl` to not reset, so I changed it to `default_url + _type=#{types}`. I tried doing `+= _types=#{types}`, but that also caused it not to reset either. 

# Testing guidance
Run the `rspec` tests and ensure that they all pass. 

Start up the test kit with `docker-compose up --build.` Try to run the non-measure-specific import test twice. For the first run, use `http://deqm_test_server:3000/4_0_1` for the URL and `Patient` for the type. Check that this passes. Then, use `http://deqm_test_server:3000/4_0_1` for the URL again and `Procedure` for the type. Check that this passes and that the `valueUrl` in the http request ends with `$export?_type=Procedure` and not `$export?_type=Patient_type=Procedure`. You can also try running without specifying any types. This will time out, but you can check that a 500 error does not get thrown right away (it used to throw 500 because it would tack on `"type="`).

### DIFF
--- a/lib/deqm_test_kit/bulk_import.rb
+++ b/lib/deqm_test_kit/bulk_import.rb
@@ -10,34 +10,30 @@ module DEQMTestKit
     description 'Ensure the fhir server can accept bulk data import requests in the non-measure-specific case'
 
     default_url = 'https://bulk-data.smarthealthit.org/eyJlcnIiOiIiLCJwYWdlIjoxMDAwMCwiZHVyIjoxMCwidGx0IjoxNSwibSI6MSwic3R1IjozLCJkZWwiOjB9/fhir/$export'
-    params = {
-      resourceType: 'Parameters',
-      parameter: [
-        {
-          name: 'exportUrl',
-          valueUrl: default_url
-        }
-      ]
-    }
+
     fhir_client do
       url :url
     end
+    # rubocop:disable Metrics/BlockLength
     test do
       title 'Ensure data can be accepted'
       id 'bulk-import-01'
       description 'POST to $import returns 202 response, bulk status endpoint returns 200 response'
 
       input :types, optional: true,
-                  description: 'string of comma-delimited FHIR resource types'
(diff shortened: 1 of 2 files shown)


lib/deqm_test_kit/bulk_import.rb
```diff
@@ -10,34 +10,30 @@ class BulkImport < Inferno::TestGroup
...
+        parameter: [
+          {
+            name: 'exportUrl',
+            valueUrl: default_url
+          }
+        ]
+      }
       run do
-        if types.length > 0
-          puts 'AAAAAAAAAAAA'
```
mgramigna: can we leave this line in actually

sarahmcdougall: stop creeping on my commits mgramigna

lmd59 (approved):
lgtm!

PR #27, opened 2022-03-04
Radio button input for measures

sarahmcdougall:
# Summary
This PR implements radio button selection for measure IDs in the test kit.

## New behavior
Previously, we had to hard-code the EXM130 measure for all the tests. The Inferno-core 0.1.X update adds the ability to specify a radio button input. Now, when running the tests, the user will be able to select a measure from the list of options. The radio button input defaults to the EXM130 measure.

## Code changes
For each of the tests that require a measure ID as input, the measure ID is now specified to be of type `radio`, is defaulted to have the EXM130 measure ID for the default value, and contains value options for the following measures: EXM104, EXM105, EXM124, EXM125, and EXM130. The fixture `measureRadioButton.json` contains the label/value pairs for the radio buttons.

The code for the radio button inputs was inspired by some of the code additions in this [Inferno-Core PR](https://github.com/inferno-framework/inferno-core/pull/70).

Some small changes were also made to the care gaps tests to reflect recent updates in `deqm-test-server`.

# Testing guidance 
All tests can be run using the rspec command. To run the application, make sure you have docker running, then run the commands `docker-compose pull` followed by `docker-compose up --build`. Navigate to `http://localhost:4567 `and select the test suite. 

Try running the tests with different measures. To access some of the newly available measure options, you may need to run `docker-compose up` and then POST the desired measure to `http://localhost:3000/4_0_1/`. Otherwise, the measure availability will fail, and all other tests that require the measure will fail.

Note that some tests are expected to fail right now, such as the `evaluate-measure` subject-list test and the bulk submit data test. Also note that some of the care gaps tests may fail and throw a 400 error if the patient ID format is not `Patient/<id>` or `Group/<id>`.

### DIFF
--- a/lib/deqm_test_kit/measure_availability.rb
+++ b/lib/deqm_test_kit/measure_availability.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'json'
+
 module DEQMTestKit
   # MeasureAvailability test group ensures selected measures are available on the fhir server
   class MeasureAvailability < Inferno::TestGroup
@@ -10,6 +12,8 @@ module DEQMTestKit
     fhir_client do
       url :url
     end
+
+    measure_options = JSON.parse(File.read('./lib/fixtures/measureAvailabilityRadioButton.json'))
     # rubocop:disable Metrics/BlockLength
     test do
       title 'Measure can be found'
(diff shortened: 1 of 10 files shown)


lib/deqm_test_kit/measure_availability.rb
```diff
@@ -10,20 +10,42 @@ class MeasureAvailability < Inferno::TestGroup
...
-
-    MEASURES = ['EXM130|7.3.000', 'EXM125|7.3.000'].freeze
-
+    # rubocop:disable Metrics/BlockLength
     test do
       title 'Measure can be found'
       id 'measure-availability-01'
       description 'Selected measure with matching id is available on the server and a valid json object'
       makes_request :measure_search
+      input :selected_measure_id, type: 'radio', optional: false,
```
natjoe4, line 19: can we use the same JSON.parse strategy here instead of listing out all the measure info?

natjoe4 (changes requested):
Just a few small comments here. Functionality works great!

PR #29, opened 2022-03-28
Inferno Core Upgrade - 0.2.0

sarahmcdougall:
# Summary
This PR upgrades inferno-core to 0.2.0.

## New behavior
The new upgrade includes:
* UI updates
* Fixes bug which prevented the running of individual tests
* Added ability to host inferno at something other than the root path
* Added a report view
* Support for defining/selecting preset inputs

## Code changes
Updated the deqm-test-kit `gemspec` to upgrade inferno-core to 0.2.0.

The `config.ru` file was replaced with [this updated version](https://github.com/inferno-framework/inferno-template/blob/main/config.ru) due to a breaking change.

# Testing guidance
Run the rspec tests with `rspec` and check that they pass.

Start up the test kit with `docker-compose pull`, followed by `docker-compose up —build`. Try running all the tests and make sure everything works as intended. Note that some of the tests are currently expected to fail (ex. bulk submit data test). These failures should be unrelated to the upgrade.

Try running individual tests and check that there are no bugs (should be fixed with the upgrade).


natjoe4 (approved):
Nice work :)

PR #30, opened 2022-03-29
Add tests for program, practitioner, organization to the care gaps tests

lmd59:
# Summary
Adds two new tests to the care gap suite to ensure the support of program and of practitioner w/ organization parameters.

## New behavior
The test kit will now run tests to check that the server under test can appropriately handle a `program={name}` parameter instead of a measure parameter, and `organization=Organization/{id}` w/ `practitioner=Practitioner/{id}` parameters instead of subject parameter.

## Code changes
- Added tests for program and practitioner/organization parameters for the $care-gaps operation
- Added rspec testing for new tests

# Testing guidance
(may need to `docker-compose pull` and/or `bundle install` if the below don't work, depending on how up to date your system already is)
- Run rspec tests using `rspec`
- Run `docker-compose up --build`
- After the server starts up, these are the following insomnia/postman requests required to run all care-gaps tests:
`PUT` to `http://localhost:3000/4_0_1/Group/EXM130-patients`, with the following body
```
{ 
  "resourceType": "Group", 
  "id": "EXM130-patients", 
  "type": "person", 
  "actual": "true", 
  "member": [
    {
      "entity": { 
        "reference": "Patient/denom-EXM130" 
      } 
    }, 
    {
      "entity": {
        "reference": "Patient/numer-EXM130"
      }
    }
  ] 
}
```
`PUT` to `http://localhost:3000/4_0_1/Organization/1`, with the following body
```
{
"resourceType": "Organization",
"id": "1"
}
```
`PUT` to `http://localhost:3000/4_0_1/Practitioner/1`, with the following body
```
{
"resourceType": "Practitioner",
"id": "1"
}
```
`PUT` to `http://localhost:3000/4_0_1/Patient/denom-EXM130`, with the following body
```
{
	"resourceType": "Patient",
	"id": "denom-EXM130",
	"managingOrganization": {
		"reference": "Organization/1"
	},
	"generalPractitioner": {
		"reference": "Practitioner/1"
	}
	
}
```
- Run the care-gaps suite using the following parameters:
Screen Shot 2022-03-29 at 10 55 49 AM
- All care gaps tests should pass

### DIFF
--- a/lib/deqm_test_kit/care_gaps.rb
+++ b/lib/deqm_test_kit/care_gaps.rb
@@ -153,6 +153,44 @@ module DEQMTestKit
         assert(resource.issue[0].severity == 'error')
       end
     end
+    test do
+      title 'Check $care-gaps proper calculation with practitioner and organization parameters'
+      id 'care-gaps-08'
+      description 'Server should properly return a gaps report for request with practitioner and organization'
+      input :measure_id, measure_id_args
+      input :period_start, default: '2019-01-01'
+      input :period_end, default: '2019-12-31'
+      input :practitioner_id
+      input :org_id
+
+      run do
+        params = "measureId=#{measure_id}&periodStart=#{period_start}&periodEnd=#{period_end}"\
+                 "&status=open-gap&practitioner=Practitioner/#{practitioner_id}&organization=Organization/#{org_id}"
+        fhir_operation("/Measure/$care-gaps?#{params}")
+
+        assert_response_status(200)
+        assert_resource_type(:parameters)
+        assert_valid_json(response[:body])
+      end
+    end
+    test do
+      title 'Check $care-gaps proper calculation with program parameter'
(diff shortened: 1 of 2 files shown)


sarahmcdougall (commented):
Great work on this - all the care gaps tests and rspec tests pass!

Would it make sense to include a test that handles a request where an organization is provided but not a practitioner? While the organization + practitioner test implicitly covers this, it might be worth adding a test with just the organization so that we are fully aligning with the spec (providing a test for subject OR organization OR organization and practitioner). On the other hand, I don't want to clutter the test kit with too many tests. Thoughts?

lmd59:
> 

Good thoughts! Sounds like we'll stick with fewer tests for now. Can always change our mind later if we find a reason for the organization only test! :-)

sarahmcdougall (approved):

PR #35, opened 2022-06-23
Added FHIR Query from Data Requirements Suite

natjoe4:
# Summary
Added a test suite that tests a servers capability to handle `GET` requests that result from the DataRequirements returned from a `$data-requirements` request

## New behavior
A new suite that takes a `data_requirements_server_url` as input and runs a `$data-requirements` operation on that server. Then, parses the returned `DataRequirements` and forms `GET` requests to the server under test based on the resourceTypes and filters listed in the DataRequirements. The suite ensures that these `GET` requests always receive `200OK` responses and valid JSON bodies.

## Code changes
 - Added `fhir_queries.rb` and `fhir_queries_spec.rb`
 - Bumped along the spec files indices for the suites I wanted to appear after `FHIR Queries`

# Testing guidance
 - `rspec`
 - Run `docker-compose pull`
 - Run `docker-compose up --build`
 - `POST` the EXM-130 bundle to `http://localhost:3000/4_0_1/`
 - Run the new `FHIR Queries` suite
 - Put in `http://deqm_test_server:3000/4_0_1` for both url and data_requirements_server_url
 - There's only one test in there, and it should pass!
 - Run all the test suites and make sure responses match what we're used to

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,8 @@
 version: "3"
 services:
   inferno:
+    stdin_open: true
+    tty: true
     build:
       context: ./
     volumes:
@@ -9,10 +11,10 @@ services:
       - "4567:4567"
     depends_on:
       - validator_service
-  cqf_ruler:
-    image: mitrehealthdocker/cqf-ruler-preloaded:0.6.0.no-vs
-    ports:
-      - "8080:8080"
+  # cqf_ruler:
+  #   image: mitrehealthdocker/cqf-ruler-preloaded:0.6.0.no-vs
+  #   ports:
+  #     - "8080:8080"
   validator_service:
     image: infernocommunity/fhir-validator-service:latest
     # Update this path to match your directory structure
@@ -40,6 +42,8 @@ services:
       - ./data/redis:/data
     command: redis-server --appendonly yes
   worker:
+    stdin_open: true
+    tty: true
     build:
       context: ./
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -40,8 +40,6 @@ services:
       - ./data/redis:/data
     command: redis-server --appendonly yes
   worker:
-    stdin_open: true
-    tty: true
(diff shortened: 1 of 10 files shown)


spec/deqm_test_kit/fhir_queries_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  def run(runnable, inputs = {})
+    test_run_params = { test_session_id: test_session.id }.merge(runnable.reference_hash)
+    test_run = Inferno::Repositories::TestRuns.new.create(test_run_params)
+    inputs.each do |name, value|
+      session_data_repo.save(test_session_id: test_session.id, name: name, value: value, type: 'text')
+    end
+    Inferno::TestRunner.new(test_session: test_session, test_run: test_run).run(runnable)
+  end
+
+  describe 'data requirements matches embedded results test' do
```
sarahmcdougall, line 25: Should probably change this description since we are not comparing against the embedded client

docker-compose.yml
```diff
@@ -9,10 +11,10 @@ services:
       - "4567:4567"
     depends_on:
       - validator_service
-  cqf_ruler:
-    image: mitrehealthdocker/cqf-ruler-preloaded:0.6.0.no-vs
-    ports:
-      - "8080:8080"
+  # cqf_ruler:
```
sarahmcdougall, line 14: Was cqf-ruler commented out for testing or do we want to keep it commented out for when this is used for testing at the Connectathon?

sarahmcdougall (changes requested):
Tested with both EXM130 and EXM124. Functionality looks good and the `rspec` tests all pass. All the search sets return 0 except for the Patient search set, but I think that's consistent with what resources live on the server after POSTing the measure bundles. Just have some small documentation changes, but otherwise we should be in good shape for the Connectathon!

sarahmcdougall (approved):

PR #36, opened 2022-07-11
FHIR Query Pattern Toggle 

natjoe4:
# Summary
Adds a true/false toggle for option to use the FHIR query pattern extension when running the FHIR Queries suite

## New behavior
New option under FHIR Queries suite to use FHIR query pattern extension. When set to true, the suite will execute FHIR queries using the url in the extension rather than the filters on the data requirement. Currently strips off the {{context.patientId}} substring included in the extension 

## Code changes
 - Added new true/false radio in FHIR Queries suite
 - Added functionality to parse fhir query extension into params object
 - Updated FHIR Queries suite logic to pull params from fqm extension when toggle set to true
 - Added testing 
 
# Testing guidance
- Run tests with `rspec`
- Boot up inferno with `docker compose up --build`
- Select FHIR Queries suite
- Run all tests, leaving `Use FHIR Query Pattern` set to false and `http://deqm_test_server:3000/4_0_1` for both `url` and `data_requirements_server_url`. Ensure all pass
- Repeat with `Use FHIR Query Pattern` set to true.
- The test should fail due to some issues with the test server's processing of the `date` parameter, but ensure that all requests that do not use the date parameter succeed with 200

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,8 @@
 version: "3"
 services:
   inferno:
+    stdin_open: true
+    tty: true
     build:
       context: ./
     volumes:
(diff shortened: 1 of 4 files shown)


sarahmcdougall (changes requested):
Tested with EXM130 and EXM124. I just have a few suggestions regarding formatting but otherwise looks good - thoroughly unit tested and the new radio button is labeled well.

When I tested, the only failures I got were for `ServiceRequest` resources, which is odd. I saw that the error was regarding the date param but there are other resources with dates and they return 200. I wonder if there's something else going on with the `ServiceRequest`s that's causing the failure? I couldn't seem to narrow it down when I was testing.

sarahmcdougall (approved):

PR #40, opened 2022-08-03
Stop Repeating Ourselves (in assertions and requests)

natjoe4:
# Summary
Broke out repeated assertion and requests structure code into reusable blocks as suggested by Inferno team

## New behavior
None

## Code changes
Added a new helpers module to practically every test suite containing functions that make requests and assert successes or failures based on the outcome of those requests. Greatly cuts down on repeated code

# Testing guidance
We want no changed behavior here, so fire up the test kit and run things as usual to make sure everything looks correct. Also run rspec tests

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,7 +4,7 @@ AllCops:
 Metrics/BlockLength:
   Enabled: true
   Exclude:
-    - 'spec/**/*'
+    - "spec/**/*"
 
 Metrics/PerceivedComplexity:
   Max: 10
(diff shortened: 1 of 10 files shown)


hossenlopp (approved):
Clever use of adding to the inferno DSL. Will have to make sure that updates to inferno don't add something same/similar. Although this implementation should override theirs.

Cleans up the tests a whole ton!

elsaperelli (approved):
Great use of helpers! Looks good to me.

PR #42, opened 2022-12-05
Inferno Template Update

sarahmcdougall:
# Summary
Updates ruby version to 3.1.2, inferno-core version to 0.4.5, and updates the Inferno template according to these [outlined instructions](https://github.com/inferno-framework/inferno-template/releases/tag/v0.0.3).

## New behavior
No new behavior is directly introduced in this PR. See the [inferno-core CHANGELOG.md](https://github.com/inferno-framework/inferno-core/blob/main/CHANGELOG.md) to see new behavior introduced by the inferno-core upgrade from 0.3.5 to 0.4.5.

## Code changes
Most of the code changes reflected in this PR come from applying the Inferno template update and from running `rubocop` after updating the ruby version to 3.1.2.

Some changes to pay attention to:
* The changes in `docker-compose.yml`/`docker-compose.background.yml`
* The use of the double-splat (**) operator for specifying `measure_id_args` as an input in the tests that require a `measure_id` to be specified. Without this operator, `inferno-core` mistakes the `measure_id_args` for another input name, which must be a symbol or a string (`measure_id_args` is a hash). Using the double-splat operator allows `inferno-core` to match `measure_id_args` to the hash argument in its input-gathering function.
* In `lib/utils/data_requirements_utils.rb` - function `get_data_requirements_queries()`’s parameter `include_patient` no longer has a default value - the inclusion of the default value was causing the corresponding `rspec` tests to fail (saying that 2 args were received when only 1 was expected). Changing `include_patient: false` to `include_patient = false` fixed the issue but raised rubocop issues. I figured we can remove the default value since we only use the function in a few places within the code.

Most of the remaining code changes are a result of applying the template patch or running rubocop.

# Testing guidance
* Run `rspec` tests
* Run in Docker with `docker compose pull` followed by `docker compose up --build`
* If you are running into hanging tests in the test kit, try running `docker-compose run inferno bundle exec rake db:migrate`
* Inspect the [commit that applies the patch](https://github.com/projecttacoma/deqm-test-kit/commit/aebae58ebf23af6d7aa0715145f955fba047c568) to ensure that all the changes are accurate in regard to deqm-test-kit


Gemfile.lock
```diff
@@ -119,155 +124,162 @@ GEM
...
+    unicode-display_width (2.3.0)
+    webmock (3.18.1)
       addressable (>= 2.8.0)
       crack (>= 0.3.2)
       hashdiff (>= 0.4.0, < 2.0.0)
-    zeitwerk (2.6.0)
+    zeitwerk (2.6.6)
 
 PLATFORMS
-  aarch64-linux
```
elsaperelli, line 263: When I run `bundle install`, `ruby` is added here.

docker-compose.background.yml
```diff
@@ -0,0 +1,59 @@
+version: '3'
+services:
+  validator_service:
+    image: infernocommunity/fhir-validator-service
+    # Update this path to match your directory structure
```
natjoe4, line 5:
```suggestion
```

elsaperelli (changes requested):
Nice work! The test-kit still works I just have three things:

1. We discussed this but it seems like the Gemfile.lock may be missing some dependencies. When I run `bundle install`, it adds some and at first I thought that was an issue but I see that they are present in https://github.com/projecttacoma/deqm-test-kit/pull/42/commits/aebae58ebf23af6d7aa0715145f955fba047c568 I am happy to help work through this

2. I am aware that rubocop made a lot of changes in some of the files but I was curious about some string formatting changes that didn't seem to make sense. They add had to do with escaping a special character in the description string. Seems like everything works regardless of that but it's weird so I pointed one out anyways. 

3. Nat and I found a bug that may be out of scope for this task but I thought I would at least bring it up. When I run the test-kit with `Patient` in the types field, everything works. However, when I leave that field blank, I get the following error: 

image

Nat thinks this may have to do with how we handle empty input.

lib/deqm_test_kit/care_gaps.rb
```diff
@@ -111,8 +111,8 @@ def care_gaps_assert_failure(params, expected_status: 400)
       title 'Check $care-gaps returns a BadRequest error for invalid subject format'
       id 'care-gaps-05'
       description %(Server should not perform calculation and return a 400 response code
-    when both the subject query parameter is not of the form Patient/\<id\> or Group/\<id\>.)
-      input :measure_id, measure_id_args
+    when both the subject query parameter is not of the form Patient/<id> or Group/<id>.)
+      input :measure_id, **measure_id_args
```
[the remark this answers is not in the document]
sarahmcdougall: Screen Shot 2022-12-06 at 12 23 17 PM
Weird, I'll look into this - might be related to this dependency (that should have fixed the issue but maybe not?)

natjoe4 (changes requested):
Looks good and functions as normal! Only one small omission and looks like the README ruby dependency should be bumped to 3.1.2!

Gemfile.lock
sarahmcdougall: Added the ruby platform back in https://github.com/projecttacoma/deqm-test-kit/pull/42/commits/45db51985c701615d7eb2c0c42c32b1cc5e38359

My thought process here: Based on the bundler docs, `PLATFORMS` represents "The ruby platform the Gemfile.lock was generated against. If any dependencies in the Gemfile specify a platform they they will only be included in that Gemfile.lock when the lock file is generated on that platform (e.g. through an install)." So, when I did an install it removed the `ruby` platform, thus removing the dependencies that you were getting when you did an install.

To bring the ruby platform back, I ran `bundle lock --add-platform ruby`. Pretty sure this updated the dependencies automatically, or maybe I had to run `bundle update` after - either way, it added those dependencies back in.

Based on this (probably not super reliable but idk)[ source](https://www.moncefbelyamani.com/understanding-the-gemfile-lock-file/ ), “Before, when projects were bundled with Bundler version 2.1.x, the platforms section in Gemfile.lock would automatically include the ruby platform, along with the current operating system. This allowed Ruby apps to be deployed on Linux servers. But since version 2.2, the ruby platform is no longer included by default.”

So, this might just be a result of the bundler version that I'm using. I did not find a ton of information about the ruby platform as a whole - not enough to determine if we actually need to include it for this project - but the [inferno-core template's lock file ](https://github.com/inferno-framework/inferno-template/blob/main/Gemfile.lock) and the [us-core-test-kit's lock file](https://github.com/inferno-framework/us-core-test-kit/blob/main/Gemfile.lock) (another test kit that uses the template) both include the ruby platform, so I'm going to use those as precedent to say we should just keep the ruby platform in for now because it doesn't hurt

lib/deqm_test_kit/bulk_import.rb
```diff
@@ -8,7 +8,7 @@ class BulkImport < Inferno::TestGroup
     id 'bulk_import'
     title 'Non-Measure-Specific Bulk Import'
     description %(
-        This test inspects the response to POST \[base\]/$import and GET \[bulk status endpoint\]
+        This test inspects the response to POST [base]/$import and GET [bulk status endpoint]
```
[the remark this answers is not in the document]
sarahmcdougall: Updated in https://github.com/projecttacoma/deqm-test-kit/pull/42/commits/5bf57455045e00ab7780441e03b52642b71836b5

Based on my ruby research, there's literally a million and one ways to define a string in ruby. This way `%()` is just one of them. If you specify a multi-line string with double quotes, then you can safely do interpolation and character escaping. I just changed these problematic strings to use double quotes for now - I didn't change the other ones to use this syntax but we can do that in the future if we really want to

Source for all the different ways to create strings: https://www.designcise.com/web/tutorial/how-to-create-multiline-strings-in-ruby

lib/deqm_test_kit/bulk_submit_data.rb
```diff
@@ -10,7 +10,7 @@ class BulkSubmitData < Inferno::TestGroup
     id 'bulk_submit_data'
     title 'Bulk Submit Data'
     description %(
```
elsaperelli, line 12:
```suggestion
    description "
```

lib/deqm_test_kit/bulk_submit_data.rb
```diff
@@ -10,7 +10,7 @@ class BulkSubmitData < Inferno::TestGroup
...
-      This test inspects the response to POST \[base\]/$submit-data and GET \[bulk status endpoint\]
+      This test inspects the response to POST [base]/$submit-data and GET [bulk status endpoint]
       to ensure that the FHIR server can accept bulk data import requests when a measure
       is specified
     )
```
elsaperelli, line 16:
```suggestion
      This test inspects the response to POST [base]/$submit-data and GET [bulk status endpoint]
      to ensure that the FHIR server can accept bulk data import requests when a measure
      is specified
    "
```

elsaperelli (changes requested):
Just one description string format change and then all is good to go! One question I have is- do you want the descriptions to all be formatted the same way for consistency? Not sure how much this matters for this repo, I just noticed that one use single quotes, some use double, and some use %().

sarahmcdougall:
> One question I have is- do you want the descriptions to all be formatted the same way for consistency?

I think it would be nice but not necessarily something that needs to get done in this PR since this PR is just supposed to apply the patch and update the inferno-core/ruby versions (really didn't think it would end up touching almost every file in the process haha). Previously, the percent-sign format was used for all multi-line descriptions and I think everything else was either single/double quotes, so at some point we should be consistent but as long as rubocop is happy I'm happy

elsaperelli (approved):
Nice!

PR #50, opened 2023-04-26
Update data requirements to not require period parameters and update …

lmd59:
…test data for ecqm-content-r4

# Summary
Updates the data-requirements tests to make periodStart and periodEnd not required, and updates the testing data options with measures that reflect the ids found in ecqm-content-r4 so that it aligns with expected database content for the deqm-test-server

## New behavior
The test kit will now run tests to check that the server under test can appropriately handle a `program={name}` parameter instead of a measure parameter, and `organization=Organization/{id}` w/ `practitioner=Practitioner/{id}` parameters instead of subject parameter.

## Code changes
- Added tests for program and practitioner/organization parameters for the $care-gaps operation
- Added rspec testing for new tests

# Testing guidance
(may need to `docker compose pull` and/or `bundle install` if the below don't work, depending on how up to date your system already is)
- Run rspec tests using `rspec`
- Run `docker compose up --build` or local start with `ASYNC_JOBS=false bundle exec puma`
- After the server starts up, these are the following insomnia/postman requests required to run all care-gaps tests:
`PUT` to `http://localhost:3000/4_0_1/Group/Cervical-patients`, with the following body
```
{ 
  "resourceType": "Group", 
  "id": "Cervical-patients", 
  "type": "person", 
  "actual": "true", 
  "member": [
    {
      "entity": { 
        "reference": "Patient/denom-EXM124" 
      } 
    }, 
    {
      "entity": {
        "reference": "Patient/numer-EXM124"
      }
    }
  ] 
}
```
`PUT` to `http://localhost:3000/4_0_1/Organization/1`, with the following body
```
{
"resourceType": "Organization",
"id": "1"
}
```
`PUT` to `http://localhost:3000/4_0_1/Practitioner/1`, with the following body
```
{
"resourceType": "Practitioner",
"id": "1"
}
```
`PUT` to `http://localhost:3000/4_0_1/Patient/denom-EXM124`, with the following body
```
{
	"resourceType": "Patient",
	"id": "denom-EXM124",
	"managingOrganization": {
		"reference": "Organization/1"
	},
	"generalPractitioner": {
		"reference": "Practitioner/1"
	}
	
}
```
- Run the care-gaps suite using the following parameters:

- All tests should pass except Bulk Data

### DIFF
--- a/lib/deqm_test_kit/data_requirements.rb
+++ b/lib/deqm_test_kit/data_requirements.rb
@@ -22,7 +22,8 @@ module DEQMTestKit
     end
 
     measure_options = JSON.parse(File.read('./lib/fixtures/measureRadioButton.json'))
-    measure_id_args = { type: 'radio', optional: false, default: 'measure-EXM130-7.3.000', options: measure_options }
+    measure_id_args = { type: 'radio', optional: false, default: 'ColorectalCancerScreeningsFHIR',
+                        options: measure_options }
 
     PARAMS = {
       resourceType: 'Parameters',
@@ -55,7 +56,7 @@ module DEQMTestKit
         measure_version = resource.version
 
         # Run our data requirements operation on the test client server
-        fhir_operation("Measure/#{measure_id}/$data-requirements?periodEnd=2019-12-31&periodStart=2019-01-01",
+        fhir_operation("Measure/#{measure_id}/$data-requirements",
                        body: PARAMS, name: :data_requirements)
         assert_response_status(200)
         assert_resource_type(:library)
@@ -72,7 +73,7 @@ module DEQMTestKit
 
         # Run data requirements operation on reference server
         fhir_operation(
-          "Measure/#{reference_measure_id}/$data-requirements?periodEnd=2019-12-31&periodStart=2019-01-01",
+          "Measure/#{reference_measure_id}/$data-requirements",
           body: PARAMS,
           client: :dr_reference_client
         )
@@ -97,15 +98,19 @@ module DEQMTestKit
     end
 
     test do
+      optional
       include DataRequirementsHelpers
-      title 'Check data requirements returns 400 for missing parameters'
+      title 'Data requirements supports optional parameters periodStart and periodEnd'
       id 'data-requirements-02'
-      description 'Data requirements returns 400 when periodStart and periodEnd parameters are omitted'
+      description 'Data requirements returns 200 when periodStart and periodEnd parameters are included'
       input :measure_id, **measure_id_args
       run do
         # Run our data requirements operation on the test client server
-        fhir_operation("Measure/#{measure_id}/$data-requirements", body: PARAMS)
-        assert_dr_failure
+        fhir_operation("Measure/#{measure_id}/$data-requirements?periodEnd=2019-12-31&periodStart=2019-01-01",
+                       body: PARAMS)
+        assert_response_status(200)
+        assert_resource_type(:library)
+        assert_valid_json(response[:body])
       end
     end
 
--- a/lib/deqm_test_kit/evaluate_measure.rb
+++ b/lib/deqm_test_kit/evaluate_measure.rb
@@ -27,7 +27,8 @@ module DEQMTestKit
     end
 
     measure_options = JSON.parse(File.read('./lib/fixtures/measureRadioButton.json'))
-    measure_id_args = { type: 'radio', optional: false, default: 'measure-EXM130-7.3.000', options: measure_options }
+    measure_id_args = { type: 'radio', optional: false, default: 'ColorectalCancerScreeningsFHIR',
+                        options: measure_options }
 
     INVALID_MEASURE_ID = 'INVALID_MEASURE_ID'
     INVALID_PATIENT_ID = 'INVALID_PATIENT_ID'
--- a/lib/deqm_test_kit/fhir_queries.rb
+++ b/lib/deqm_test_kit/fhir_queries.rb
@@ -40,7 +40,7 @@ module DEQMTestKit
     title 'FHIR Queries'
     description 'Ensure FHIR server can handle queries resulting from $data-requirements operation'
     measure_options = JSON.parse(File.read('./lib/fixtures/measureRadioButton.json'))
-    measure_id_args = { type: 'radio', optional: false, default: 'measure-EXM130-7.3.000',
+    measure_id_args = { type: 'radio', optional: false, default: 'ColorectalCancerScreeningsFHIR',
                         options: measure_options }
 
     use_fqp_extension_args = {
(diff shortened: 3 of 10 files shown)


lib/deqm_test_kit/evaluate_measure.rb
```diff
@@ -27,7 +27,8 @@ def measure_evaluation_assert_failure(params, measure_id, expected_status: 400)
     end
 
     measure_options = JSON.parse(File.read('./lib/fixtures/measureRadioButton.json'))
-    measure_id_args = { type: 'radio', optional: false, default: 'measure-EXM130-7.3.000', options: measure_options }
+    measure_id_args = { type: 'radio', optional: false, default: 'ColorectalCancerScreeningsFHIR',
+                        options: measure_options }
```
elsaperelli, line 31:
```suggestion
    measure_id_args = { type: 'radio', optional: false, default: 'ColorectalCancerScreeningsFHIR',
                        options: measure_options, title: 'Measure ID' }
```
I like this addition to the other suites!!

lib/deqm_test_kit/fhir_queries.rb
```diff
@@ -40,7 +40,7 @@ def fhir_queries_assertions(queries)
     title 'FHIR Queries'
     description 'Ensure FHIR server can handle queries resulting from $data-requirements operation'
     measure_options = JSON.parse(File.read('./lib/fixtures/measureRadioButton.json'))
-    measure_id_args = { type: 'radio', optional: false, default: 'measure-EXM130-7.3.000',
+    measure_id_args = { type: 'radio', optional: false, default: 'ColorectalCancerScreeningsFHIR',
                         options: measure_options }
```
elsaperelli, line 44:
```suggestion
                        options: measure_options, title: 'Measure ID' }
```

lib/deqm_test_kit/submit_data.rb
```diff
@@ -37,7 +37,8 @@ def create_measure_report(measure_url, period_start, period_end)
     description 'Ensure fhir server can receive data via the $submit-data operation'
     custom_headers = { 'X-Provenance': '{"resourceType": "Provenance", "agent": ["test-agent"]}' }
     measure_options = JSON.parse(File.read('./lib/fixtures/measureRadioButton.json'))
-    measure_id_args = { type: 'radio', optional: false, default: 'measure-EXM130-7.3.000', options: measure_options }
+    measure_id_args = { type: 'radio', optional: false, default: 'ColorectalCancerScreeningsFHIR',
+                        options: measure_options }
```
elsaperelli, line 41:
```suggestion
                        options: measure_options, title: 'Measure ID' }
```

lib/deqm_test_kit/data_requirements.rb
```diff
@@ -22,7 +22,8 @@ def assert_dr_failure(expected_status: 400)
     end
 
     measure_options = JSON.parse(File.read('./lib/fixtures/measureRadioButton.json'))
-    measure_id_args = { type: 'radio', optional: false, default: 'measure-EXM130-7.3.000', options: measure_options }
+    measure_id_args = { type: 'radio', optional: false, default: 'ColorectalCancerScreeningsFHIR',
+                        options: measure_options }
```
elsaperelli, line 26:
```suggestion
                        options: measure_options, title: 'Measure ID' }
```

lib/deqm_test_kit/measure_availability.rb
```diff
@@ -22,7 +22,8 @@ def measure_availability_assert_success(measure_identifier, measure_version)
     end
 
     measure_options = JSON.parse(File.read('./lib/fixtures/measureAvailabilityRadioButton.json'))
-    measure_id_args = { type: 'radio', optional: false, default: 'EXM130|7.3.000', options: measure_options }
+    measure_id_args = { type: 'radio', optional: false, default: 'ColorectalCancerScreeningsFHIR|0.0.003',
+                        options: measure_options }
```
elsaperelli, line 26:
```suggestion
                        options: measure_options, title: 'Measure ID' }
```

elsaperelli (changes requested):
Great work with this! I think it's important to go through which test-kit suites pass every so often so thanks for stepping up! Just some small comments. In terms of the outstanding issues- I think we should create some backlog tasks to address them (if we don't already) since this is only pointed as 1 point.

sarahmcdougall:
> In terms of the outstanding issues- I think we should create some backlog tasks to address them (if we don't already) since this is only pointed as 1 point.

I added (1) updating the fqm-execution dependency to 1.1.0 for the test server, and (2) update deqm-test-kit docker image, as end of sprint tasks to my running list of EOS work. But I can add them to the backlog as sprint tasks if folks think that would add more transparency/awareness to the outstanding work that needs to be done here! Both seem pretty small in scope to me

sarahmcdougall (changes requested):
Great work tackling all these refactors to the test kit! Largely looks good - left a few in-line comments but nothing major.

In terms of the `evaluate-measure` population test that times out -- at first I thought it was directly an issue with scalable calculation because we exceed the scalable calculation threshold in deqm-test-server's `.env` (which is 100). After doing some testing in deqm-test-server directly, it looks like the issue might actually be coming from the `generalPractitioner` reference in the `denom-EXM124` fixture that you listed in the PR description. Without having that fixture in the mongo database, I am able to successfully run the test (and get a `200`) using scalable calculation. With the `denom-EXM124` fixture in the database, I run into the following error in the test server:

deqm-test-kit-eval-measure

Note the inclusion of `Patient.get [as generalPractitioner]`. Not sure the best path forward here... in my opinion this seems out of scope of this PR because the test kit logic itself is not affecting this - it's more so the data in the database and the handling through deqm-test-server and eventually cql-exec-fhir. Let me know your thoughts!

lib/deqm_test_kit/evaluate_measure.rb
lmd59: Since these are technically the measure titles, what do you think of something like:
```suggestion
    measure_id_args = { type: 'radio', optional: false, default: 'ColorectalCancerScreeningsFHIR',
                        options: measure_options, title: 'Measure Title' }
```
?

lmd59:
> Great work tackling all these refactors to the test kit! Largely looks good - left a few in-line comments but nothing major.
> 
> In terms of the `evaluate-measure` population test that times out -- at first I thought it was directly an issue with scalable calculation because we exceed the scalable calculation threshold in deqm-test-server's `.env` (which is 100). After doing some testing in deqm-test-server directly, it looks like the issue might actually be coming from the `generalPractitioner` reference in the `denom-EXM124` fixture that you listed in the PR description. Without having that fixture in the mongo database, I am able to successfully run the test (and get a `200`) using scalable calculation. With the `denom-EXM124` fixture in the database, I run into the following error in the test server:
> 
> deqm-test-kit-eval-measure
> 
> Note the inclusion of `Patient.get [as generalPractitioner]`. Not sure the best path forward here... in my opinion this seems out of scope of this PR because the test kit logic itself is not affecting this - it's more so the data in the database and the handling through deqm-test-server and eventually cql-exec-fhir. Let me know your thoughts!

I added an end of sprint activity to look into this and the bulk-submit-data issue. We can see if we can look into that further at the end of the sprint or make tickets if we feel it's ever of higher priority!

sarahmcdougall (approved):
Changes look good - thanks for documenting the end of sprint activity for this! If we don't get to it at the end of this sprint I can make a backlog task for it

elsaperelli (approved):
Lgtm!

PR #58, opened 2025-07-31
Update Test Ids

zacharyrobin:
# Summary

Updated all the id of the tests so that we no longer have to rely on numbering and making sure the spec tests are in the correct order. This helps with when we want to add a new test we can easily update the spec test without having to renumber everything else. 

Noticed while going through the Ids that we are missing tests for `measure-availability-not-found' and for `evaluate-subjectgroup-embedded-resource`. Not sure if this is by choice but happy to add them in if we desire. 

## New behavior

This update should not effect behavior at all. 

## Code changes

 Updated Test Files:

  1. `evaluate.rb`: Updated 16 test IDs 
  2. `patient_everything.rb`: Updated 3 test IDs 
  3. `measure_availability.rb`: Updated 2 test IDs
  4. `data_requirements.rb`: Updated 3 test IDs 
  5. `submit_data.rb`: Updated 3 test IDs 
  6. `fhir_queries.rb`: Updated 2 test IDs 
  7. `bulk_submit_data.rb`: Updated 1 test ID
  8. `bulk_import.rb`: Updated 1 test ID
  9. `evaluate_measure.rb`: Updated 11 test IDs
  10. `care_gaps.rb`: Updated 9 test IDs

  Updated Spec Files:

 -  All corresponding spec test files were updated to with matching ids
 -  Added a helper function in `spec_helper.rb` to handle the prefix that is prepended (ie `deqm_v500`) 

# Testing guidance

- Run the spec tests: `bundle exec rspec` <- They should all pass
- `ASYNC_JOBS=false bundle exec puma` to run the test kit locally, all tests should run as normal

### DIFF
--- a/lib/deqm_test_kit/bulk_import.rb
+++ b/lib/deqm_test_kit/bulk_import.rb
@@ -19,7 +19,7 @@ module DEQMTestKit
     # rubocop:disable Metrics/BlockLength
     test do
       title 'Ensure FHIR server can accept bulk data import requests'
-      id 'bulk-import-01'
+      id 'bulk-import-accepts-import-requests'
       description %(POST request to $import returns 202 response,
       GET request to bulk status endpoint returns 200 response)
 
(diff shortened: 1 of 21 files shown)


lmd59 (commented):
Good updates and goo descriptions! Had a few recommendations that might make the ids a bit clearer if you're looking at them quickly. Let me know if you think any of these recommendations don't make sense!

PR #59, opened 2025-08-12
Custom Input Field for MeasureID

zacharyrobin:
# Summary
Added option for custom text input from user when selecting a measureId. 

## New behavior
Perviously users selected a measureId from a list of radio buttons and if we wanted to add another we had to go into the code and manually change it. This feature allows for the user to select "other" as an option and then custom enter in the measureId they want in a text box. 
Screenshot 2025-08-12 at 12 54 09 PM

## Code changes
In 
`/evalaute.rb`
`/measure_availability.rb`
`/bulk_submit_data.rb`
`/care_gaps.rb`
`/data_requirements.rb`
`/fhir_queries.rb`
`/submit_data.rb`
`/evaluate_measure.rb`
- Added `custom_measure_id_args` for custom text input 
- Added helper function to decide if we are using a preset radio button or input from text field 
- ~~Moved shared inputs for the entire evaluate test group to the top to reduce redundancy~~
- Updated all tests that use `measure_ids` -> `selected_measure_ids`

In `/evalaute_spec.rb`
- Updated failing tests with missing information, this includes invalide_dates and removal of unnecessary parameters that were failing the spec test

In 
`/data_requirements_spec.rb`
`/bulk_submit_data_spec.rb`
- updated failing tests to accommodate for custom input changes 

In `measureRadioButton.json`
- Added "Other" option that the user can select for custom input

# Testing guidance
- Run test with `bundle exec rspec`
- Start up the test kit with `ASYNC_JOBS=false bundle exec puma`
- Start up test server with `npm run start` in `/deqm-test-server`
- Run the tests with both the local server and with Bellese Server both with the radio buttons and with custom input by selecting "Other" and entering in `ColorectalCancerScreeningsFHIR` or `CMS1017FHIRHHFI` respectively for example.


elsaperelli:
I noticed that `evaluate_measure.rb` and `care_gaps.rb` need the changes as well, when running the DEQM STU3 test suite, we don't have the ability to use the "Other" option and input a measureId. In addition, `fhir_queries.rb` and `data_requirements.rb` need these changes. Make sure all the test groups are able to run when clicking the "Other" option.

lib/deqm_test_kit/evaluate.rb
```diff
@@ -55,10 +61,32 @@ def validate_measure_report_fields(report) # rubocop:disable Metrics/AbcSize
...
+      optional: true,
+      title: 'Custom Measure ID',
+      description: 'If you selected "Other" above or want to provide a custom Measure ID, enter it here.'
+    }
+
+    # Shared inputs for the entire Evaluate test group
+    input :measure_id, **measure_id_args
+    input :custom_measure_id, **custom_measure_id_args
+    input :period_start, title: 'Measurement period start', default: '2026-01-01'
+    input :period_end, title: 'Measurement period end', default: '2026-12-31'
```
elsaperelli: I am a little hesitant to pull these out like this, but I am open to discussion. If the answer is just that it makes this file cleaner, I think we should move it back, but if you have other reasoning then I am curious.

I say maybe they should go back because some of these tests don't need them. For example, if the user just goes to run test 5.09 (why they would do this idk LOL, but they can), they HAVE to input a measure id in order to run it. This feels a little weird since the test doesn't use a measure id at all, since the test is checking that an error is throw when using an invalid measureId, which we provide. Again, open to discussion but I think what we had before makes more sense.

elsaperelli (changes requested):
Looking good! I just noticed that a handful of files are still missing changes so the Other input functionality doesn't work in those, but before you do that, I recommend reverting the shared input changes...I explain my reasoning in a comment. Nice job!

elsaperelli (changes requested):
One more fix- looks like the additional measure checkboxes in `evaluate.rb` have an "Other" checkbox but no text input for the user to include an other measureId. I would just add one text input for now under the assumption that the user can only include one additional "Other" measureId for the additional measures.

lib/fixtures/additionalMeasureRadioButton.json
```diff
@@ -0,0 +1,68 @@
+{
```
elsaperelli: I would call this file something like "measureCheckBoxes.json" since it's not being used for the radio button (I think).

spec/deqm_test_kit/v3.0.0/data_requirements_spec.rb
```diff
@@ -82,6 +82,7 @@ def run(runnable, inputs = {})
 
       result = run(test, url:,
                          measure_id:,
+                         custom_measure_id: nil,
```
elsaperelli:
```suggestion
```
These are not necessary, so I would just remove them (rest of the file).

spec/deqm_test_kit/v5.0.0/evaluate_spec.rb
```diff
@@ -602,7 +602,7 @@ def run(runnable, inputs = {})
       )
         .to_return(status: 400, body: error_outcome.to_json)
 
-      result = run(test, url:, measure_id:, patient_id:, period_start:)
+      result = run(test, url:, measure_id:, patient_id:, period_start:, period_end: '2019-12-31')
```
elsaperelli:
```suggestion
      result = run(test, url:, measure_id:, patient_id:, period_start:)
```
This can be removed now that the inputs were moved back

spec/deqm_test_kit/v5.0.0/evaluate_spec.rb
```diff
@@ -628,7 +628,7 @@ def run(runnable, inputs = {})
              })
         .to_return(status: 400, body: error_outcome.to_json)
 
-      result = run(test, url:, measure_id:, patient_id:, period_start:)
+      result = run(test, url:, measure_id:, patient_id:, period_start:, period_end: '2019-12-31')
```
elsaperelli:
```suggestion
      result = run(test, url:, measure_id:, patient_id:, period_start:)
```
This can be removed now that the inputs were moved back

spec/deqm_test_kit/v3.0.0/bulk_submit_data_spec.rb
```diff
@@ -39,7 +39,7 @@ def run(runnable, inputs = {})
 
       stub_request(:get, polling_url)
         .to_return(status: 200, body: resource.to_json)
-      result = run(test, url:, measure_id:, exportUrl:)
+      result = run(test, url:, measure_id:, custom_measure_id: nil, exportUrl:)
```
elsaperelli:
```suggestion
      result = run(test, url:, measure_id:, exportUrl:)
```

lib/deqm_test_kit/evaluate.rb
```diff
@@ -491,13 +525,15 @@ def validate_measure_report_fields(report) # rubocop:disable Metrics/AbcSize
       description %(Server should not perform calculation and return a 400 response code
       when the subject report type is specified but no subject has been specified in the
       query parameters.)
+
```
elsaperelli:
```suggestion
```

lib/deqm_test_kit/evaluate.rb
```diff
@@ -474,13 +506,15 @@ def validate_measure_report_fields(report) # rubocop:disable Metrics/AbcSize
       title 'Measure/[id]/$evaluate reportType=subject fails for invalid patient ID'
       id 'evaluate-measureid-path-invalid-patientid'
       description 'Request returns a 404 error when the given patient ID cannot be found'
+
```
elsaperelli:
```suggestion
```

lib/deqm_test_kit/evaluate.rb
```diff
@@ -438,7 +468,9 @@ def validate_measure_report_fields(report) # rubocop:disable Metrics/AbcSize
       title 'Measure/$evaluate reportType=subject fails for invalid patient ID'
       id 'evaluate-invalid-patientid-subject-body'
       description 'Request returns a 404 error when the given patient ID cannot be found.'
+
```
elsaperelli:
```suggestion
```

elsaperelli (changes requested):
Just a few more comments!

spec/deqm_test_kit/v3.0.0/data_requirements_spec.rb
```diff
@@ -144,6 +144,7 @@ def run(runnable, inputs = {})
...
```
elsaperelli:
```suggestion
```

spec/deqm_test_kit/v3.0.0/data_requirements_spec.rb
```diff
@@ -206,6 +207,7 @@ def run(runnable, inputs = {})
...
```
elsaperelli:
```suggestion
```

spec/deqm_test_kit/v3.0.0/data_requirements_spec.rb
```diff
@@ -254,6 +256,7 @@ def run(runnable, inputs = {})
...
```
elsaperelli:
```suggestion
```

spec/deqm_test_kit/v3.0.0/data_requirements_spec.rb
```diff
@@ -304,6 +307,7 @@ def run(runnable, inputs = {})
...
```
elsaperelli:
```suggestion
```

elsaperelli (changes requested):
A few more lines that need to be removed

elsaperelli (approved):
🥳

PR #60, opened 2025-08-14
DEQM v5 submit-data Tests

zacharyrobin:
# Summary
This PR adds a test to the DEQM v5.0.0 Test Suite for the [$submit-data operation](https://build.fhir.org/ig/HL7/davinci-deqm/OperationDefinition-submit-data.html).

## New behavior
- `$submit-data` test accepts multiple measure as comma-separated format: `http://example.org/Measure/A|1.0.0,http://example.org/Measure/B|1.2.3`
- Test for one subject with multiple measures 
- Test for two subjects with multiple measures

## Code changes
- `submit_data_v3.rb` - renamed to SubmitDataV3 so we can have two versions of SubmitData tests
- `submit_data_v5.rb` - created SubmitDataV5 test for the DEQM IG STU 5 version of $submit-data
- `v3.0.0/deqm_test_suite.rb` / `v5.0.0/deqm_test_suite.rb` - updated imports
- `v3.0.0/data_requirements_spec.rb` / `v5.0.0/evaluate_spec.rb` - remove old puts statements
- `v3.0.0/'submit_data_v3_spec.rb` - update name
- `v5.0.0/submit_data_spec.rb` - added spec test for new test
- `fixtures/submitDataResponse.json` - added example $submit-data output JSON for use in the spec tests

# Testing guidance
- `bundle exec rspec`
- `bundle exec rubocop`
- Start up the test kit with `ASYNC_JOBS=false bundle exec puma`
- Start up test server with `npm run start` in `deqm-test-server` (specifically on the branch in [this PR](https://github.com/projecttacoma/deqm-test-server/pull/169))
- Use the DEQM Test Server Local preset in the DEQM v5.0.0 test suite


lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,371 @@
...
+
+        "#{measure_url}|#{measure_version}"
+      end
+
+      # DELETEME: This was based on fqm-testify
+      def create_data_exchange_measure_report(measure_canonical, period_start, period_end, subject_id) # rubocop:disable Metrics/MethodLength
+        mr_hash = {
+          'resourceType' => 'MeasureReport',
+          'id' => SecureRandom.uuid,
+          'measure' => measure_canonical,
```
elsaperelli, line 22: This should probably be split by url and version

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,371 @@
...
+
+require 'securerandom'
+require 'json'
+
+module DEQMTestKit
+  # Perform submit data operation on test client
+  class SubmitDataV5 < Inferno::TestGroup
+    # module for shared code for $submit-data assertions and requests
+    module SubmitDataHelpers # rubocop:disable Metrics/ModuleLength
+      def canonical_measure(measure_url, measure_version)
```
elsaperelli: I don't think this function is necessary. `parse_measures` appears to take a measure url string, split it's url and version, which is then passed into this function, but then this function just puts them back together? I am going to remove this so that the `parse_measures` function handles all of the logic.

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,371 @@
...
+        validate_parameters_grouped_by_subject(params)
+
+        # Submit
+        fhir_operation('Measure/$submit-data', body: params, name: :submit_data_two_subjects)
+        assert_response_status(200)
+        assert_valid_json(response[:body]) # SUT may return OperationOutcome or similar
+      end
+    end
+
+    # DELETEME: ____ FALURE TESTS IF WE WANT THEM ____
```
elsaperelli: The DEQM IG isn't exactly clear on error cases, so I am going to delete this for now.

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,371 @@
...
+                             .uniq
+          assert mr_subjects.size == 1,
+                 "Expected exactly one unique subject reference in the Bundle, got: #{mr_subjects}"
+          mr_subjects.first
+        end
+
+        assert bundle_subjects.uniq.size == bundle_subjects.size,
+               "Expected each Bundle to represent a distinct subject, got duplicates: #{bundle_subjects}"
+      end
+    end
```
elsaperelli: Two main things here:

1. We are making our own bundles, measure reports, and parameters resources, so we don't need to validate them. I imagine this was from looking at $evaluate and other tests where the output to operations gives us resources that we want to validate, but this test we provide the input so we don't need to validate them.

2. You don't need any of these validation functions, but for validation in the future, remember that the Inferno DSL has a _lot_ of validation methods. For example, the [`resource_is_valid`](https://inferno-framework.github.io/inferno-core/docs/Inferno/DSL/FHIRValidation.html) and other [assertion methods](https://inferno-framework.github.io/inferno-core/docs/Inferno/DSL/Assertions.html#assert_valid_resource-instance_method) would be useful in these methods rather than manually checking the values or the presence of different attributes.

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,371 @@
...
+            description: 'Example: http://example.org/Measure/A|1.0.0,http://example.org/Measure/B|1.2.3 ' \
+                         '(also supports JSON array format)'
+
+      run do
+        measures = parse_measures(measures_json, measure_url, measure_version)
+
+        patient   = create_dummy_patient
+        encounter = create_dummy_encounter(patient.id)
+
+        reports   = build_measure_reports_for_subject(measures, period_start, period_end, patient.id)
```
elsaperelli: Is there a reason for this style of formatting where all of the `=` are in line? If not, I am not sure we should have it like that.

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,371 @@
...
+        # Subject 1
+        patient1   = create_dummy_patient
+        encounter1 = create_dummy_encounter(patient1.id)
+        reports1   = build_measure_reports_for_subject(measures, period_start, period_end, patient1.id)
+        bundle1    = create_submit_bundle(*reports1, patient1, encounter1)
+        validate_bundle_contains_data_collection_measure_report(bundle1)
+        validate_bundle_is_single_subject(bundle1)
+
+        # Subject 2
+        patient2   = create_dummy_patient
```
elsaperelli: We probably don't want patient1 and patient2 here to be the exact same, going to change this.

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,371 @@
...
+               "Expected each Bundle to represent a distinct subject, got duplicates: #{bundle_subjects}"
+      end
+    end
+    id 'submit_data_v5'
+    title '$submit-data-v5'
+    description 'Ensure fhir server can receive data via the $submit-data operation'
+    custom_headers = { 'X-Provenance': '{"resourceType": "Provenance", "agent": ["test-agent"]}' }
+
+    input :measure_url, title: 'Measure URL (canonical base)', type: 'text', optional: true
+    input :measure_version, title: 'Measure Version (optional)', type: 'text', optional: true
```
elsaperelli: I am going to remove these and just use the comma separated input.

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,371 @@
...
+        patient   = create_dummy_patient
+        encounter = create_dummy_encounter(patient.id)
+
+        reports   = build_measure_reports_for_subject(measures, period_start, period_end, patient.id)
+
+        bundle = create_submit_bundle(*reports, patient, encounter)
+
+        # Validate bundles
+        validate_bundle_contains_data_collection_measure_report(bundle)
+        validate_bundle_is_single_subject(bundle)
```
elsaperelli: Removing these, we created these bundles so we don't need to validate them.

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,371 @@
...
+      include SubmitDataHelpers
+      title 'Submit Data valid submission (one subject, multiple measures)'
+      id 'submit-data-valid-one-subject-multi-measures'
+      description 'Submit a Parameters resource containing a single Bundle with one Patient, one Encounter, ' \
+                  'and a data-collection MeasureReport per requested measure.'
+      makes_request :submit_data
+
+      input :measure_url
+      input :measure_version
+      input :measures_json,
```
elsaperelli: This variable name is misleading since it can either be a string or a JSON array.

elsaperelli: I am also going to pull this out of the test and into the test group so that it is required.

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,371 @@
...
+    test do
+      include SubmitDataHelpers
+      title 'Submit Data valid submission (one subject, multiple measures)'
+      id 'submit-data-valid-one-subject-multi-measures'
+      description 'Submit a Parameters resource containing a single Bundle with one Patient, one Encounter, ' \
+                  'and a data-collection MeasureReport per requested measure.'
+      makes_request :submit_data
+
+      input :measure_url
+      input :measure_version
```
elsaperelli: Removing these, we only want the comma-separated URLs with versions input.

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,371 @@
...
+      input :measures_json,
+            title: 'Measures (comma-separated URLs with versions)',
+            type: 'text',
+            optional: true,
+            description: 'Example: http://example.org/Measure/A|1.0.0,http://example.org/Measure/B|1.2.3 ' \
```
elsaperelli: I don't think we need to support it as JSON array format, just the comma-separated string is good!

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,371 @@
...
+      def wrap_bundles_in_parameters(*bundles)
+        {
+          resourceType: 'Parameters',
+          parameter: bundles.map { |b| { name: 'bundle', resource: b } }
+        }
+      end
+
+      # Accepts multiple measure reports as comma-separated format: `http://example.org/Measure/A|1.0.0,http://example.org/Measure/B|1.2.3`
+      # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
+      def parse_measures(measures_string, fallback_url, fallback_version)
```
elsaperelli: After talking with @lmd59, we decided we don't need to support this list as JSON, so we don't need this method. Just a `.split(',')` within the test should be good!

elsaperelli (changes requested):
Good start! I am going to push changes shortly, but I wanted to leave some review comments that will hopefully explain why I removed/changed things!

spec/deqm_test_kit/v3.0.0/data_requirements_spec.rb
```diff
@@ -328,7 +328,6 @@ def run(runnable, inputs = {})
              })
         .to_return(status: 200, body: test_library_response.to_json)
       result = run(test, url:, measure_id:, period_start:, period_end:)
-      puts result
```
lmd59, line 331: Was this cruft that we just noticed?

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,162 @@
...
+
+    id 'submit_data_v5'
+    title '$submit-data-v5'
+    description 'Ensure fhir server can receive data via the $submit-data operation'
+    custom_headers = { 'X-Provenance': '{"resourceType": "Provenance", "agent": ["test-agent"]}' }
+
+    input :measure_url_list,
+          title: 'Measures (comma-separated URLs with versions)',
+          type: 'text',
+          optional: true,
```
lmd59: Should this be optional? I think we should probably require it, but if we decide not to, we need to do some more handling of the case where it isn't input (i.e. `split()` fails).

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,162 @@
...
+            'profile' => ['http://hl7.org/fhir/us/davinci-deqm/StructureDefinition/datax-measurereport-deqm']
+          },
+          'extension' => [
+            {
+              'url' => 'http://hl7.org/fhir/us/davinci-deqm/StructureDefinition/extension-submitDataUpdateType',
+              'valueCode' => 'snapshot'
+            }
+          ],
+          'contained' => [
+            # TODO: Check what sort of organization we should be using here?
```
lmd59: I think this is fine for now based on our best knowledge of the spec. Can probably remove this TODO.

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,162 @@
...
+      end
+
+      def create_encounter(patient_id)
+        FHIR::Encounter.new(
+          'id' => "enc-#{SecureRandom.uuid}",
+          'identifier' => [{ 'value' => SecureRandom.uuid }],
+          'status' => 'finished',
+          'class' => { 'system' => 'http://terminology.hl7.org/CodeSystem/v3-ActCode', 'code' => 'AMB' },
+          'subject' => { 'reference' => "Patient/#{patient_id}" },
+          'period' => { 'start' => '2019-06-01', 'end' => '2019-06-01' }
```
lmd59: Just in case this function gets used for something else later, like testing calculation, it might be nice to at least differentiate the start and end of this period by a little. Suggest offset of one day.

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,162 @@
...
+        )
+      end
+
+      def create_submit_bundle(measure_reports, patient, encounter)
+        entries = measure_reports.map do |mr|
+          { 'resource' => mr.to_hash }
+        end
+        FHIR::Bundle.new(
+          'type' => 'transaction',
+          'entry' => entries.push(patient, encounter)
```
lmd59: These are pushed directly as resources. Needs to be added as entries with at least a resource field (i.e. `{"resource": <resource goes here>`})

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,162 @@
...
+          'status' => 'finished',
+          'class' => { 'system' => 'http://terminology.hl7.org/CodeSystem/v3-ActCode', 'code' => 'AMB' },
+          'subject' => { 'reference' => "Patient/#{patient_id}" },
+          'period' => { 'start' => '2019-06-01', 'end' => '2019-06-01' }
+        )
+      end
+
+      def create_submit_bundle(measure_reports, patient, encounter)
+        entries = measure_reports.map do |mr|
+          { 'resource' => mr.to_hash }
```
lmd59: We define the `resource` of the bundle entry but I also think we need to define the `request` with a request method (PUT/POST) for the entry as well

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,162 @@
...
+        measures = measure_url_list.split(',')
+
+        patient = create_patient('Patient1')
+        encounter = create_encounter(patient.id)
+
+        reports = build_measure_reports_for_subject(measures, period_start, period_end, patient.id)
+
+        bundle = create_submit_bundle(reports, patient, encounter)
+
+        # Wrap em in a Parameters resource
```
lmd59:
```suggestion
        # Wrap bundles in a Parameters resource
```
lol

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,162 @@
...
+        reports1 = build_measure_reports_for_subject(measures, period_start, period_end, patient1.id)
+        bundle1 = create_submit_bundle(reports1, patient1, encounter1)
+
+        # Subject 2
+        patient2 = create_patient('Patient2')
+        encounter2 = create_encounter(patient2.id)
+        reports2 = build_measure_reports_for_subject(measures, period_start, period_end, patient2.id)
+        bundle2 = create_submit_bundle(reports2, patient2, encounter2)
+
+        # Wrap em in a Parameters resource
```
lmd59:
```suggestion
        # Wrap bundles in a Parameters resource
```

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -0,0 +1,162 @@
...
+        bundle = create_submit_bundle(reports, patient, encounter)
+
+        # Wrap em in a Parameters resource
+        params_hash = wrap_bundles_in_parameters(bundle)
+        params = FHIR::Parameters.new(params_hash)
+
+        # $submit-data operation with passed in FHIR Parameters resource we created
+        fhir_operation('Measure/$submit-data', body: params)
+        assert_response_status(200)
+        assert_valid_json(response[:body]) # TODO: update to validate proper return once we confirm
```
lmd59: This TODO should probably still be done for this and the below test. I believe that our current plan is to return a Parameters with 1..* transaction-response bundles. Maybe just a check for a Parameters and that each included parameter is a with bundle resource with a `transaction-response` type.

lmd59 (commented):
Working on fixing deqm-test-server to test this better, but in the meantime, here are comments!
Additionally, we should probably add an intelligent preset for the Measures input, maybe:
http://ecqi.healthit.gov/ecqms/Measure/BreastCancerScreeningsFHIR|0.0.002,http://ecqi.healthit.gov/ecqms/Measure/CervicalCancerScreeningFHIR|0.0.005

lib/deqm_test_kit/submit_data_v5.rb
lmd59: Here's an example output:
```
{
  resourceType: "Parameters",
  parameter: [
    {
      name: "responseBundle",
      resource: {
        resourceType: "Bundle",
        id: "4eac5c8d-fde9-44f3-a4b2-290a8e4777e3",
        type: "transaction-response",
        link: [
          {
            relation: "self",
            url: "http://127.0.0.1:54837",
          },
        ],
        entry: [
          {
            response: {
              status: "201 Created",
              location: "4_0_1/MeasureReport/measurereport-testMeasure",
            },
          },
          {
            response: {
              status: "201 Created",
              location: "4_0_1/MeasureReport/measurereport-testMeasure2",
            },
          },
          {
            response: {
              status: "201 Created",
              location: "4_0_1/Patient/test-patient",
            },
          },
          {
            response: {
              status: "201 Created",
              location: "4_0_1/Encounter/test-encounter",
            },
          },
        ],
      },
    },
    {
      name: "responseBundle",
      resource: {
        resourceType: "Bundle",
        id: "6d88e86f-b510-4f6f-8e00-d18d6fcaa31a",
        type: "transaction-response",
        link: [
          {
            relation: "self",
            url: "http://127.0.0.1:54837",
          },
        ],
        entry: [
          {
            response: {
              status: "201 Created",
              location: "4_0_1/MeasureReport/measurereport-testMeasure-report2",
            },
          },
          {
            response: {
              status: "201 Created",
              location: "4_0_1/MeasureReport/measurereport-testMeasure2-report2",
            },
          },
          {
            response: {
              status: "201 Created",
              location: "4_0_1/Patient/test-patient2",
            },
          },
          {
            response: {
              status: "201 Created",
              location: "4_0_1/Encounter/test-encounter2",
            },
          },
        ],
      },
    },
  ],
}
```

spec/deqm_test_kit/v3.0.0/data_requirements_spec.rb
elsaperelli: Yes

lib/deqm_test_kit/submit_data_v5.rb
elsaperelli: I agree this should be required.

lib/deqm_test_kit/submit_data_v5.rb
elsaperelli: Good call!

lib/deqm_test_kit/submit_data_v5.rb
```diff
@@ -53,13 +53,18 @@ def create_encounter(patient_id)
         )
       end
 
-      def create_submit_bundle(measure_reports, patient, encounter)
+      def create_submit_bundle(measure_reports, patient, encounter) # rubocop:disable Metrics/MethodLength
         entries = measure_reports.map do |mr|
-          { 'resource' => mr.to_hash }
+          { 'resource' => mr.to_hash, 'request' => { method: 'PUT', url: "MeasureReport/#{SecureRandom.uuid}" } }
```
lmd59: Should these ids in the request ideally match the resource's id?

lmd59 (dismissed):
Looking good! One small question, but I don't think it's blocking!

lib/deqm_test_kit/submit_data_v5.rb
elsaperelli: Certainly :D

lmd59 (approved):
LGTM!

elsaperelli (approved):

PR #64, opened 2026-06-24
STU 1 Suite and initial collect-data success tests

elsaperelli:
# Summary
This PR adds a Universal Realm STU 1 Test Suite to the DEQM Test Kit that includes a basic capability statement test as well as 4 success tests for the $collect-data operation.

## New behavior
The 4 $collect-data success tests are as follows:
1. Tests GET Measure/$collect-data with one measureUrl, periodStart, periodEnd returns 200 status and FHIR Parameters resource where Parameters.parameter contains FHIR Bundle resources.
2. Tests POST Measure/$collect-data with one measureUrl, periodStart, periodEnd returns 200 status and FHIR Parameters resource where Parameters.parameter contains FHIR Bundle resources.
3. Tests GET Measure/$collect-data with two measureUrls, periodStart, periodEnd returns 200 status and FHIR Parameters resource where Parameters.parameter contains FHIR Bundle resources.
4. Tests POST Measure/$collect-data with two measureUrls, periodStart, periodEnd returns 200 status and FHIR Parameters resource where Parameters.parameter contains FHIR Bundle resources.

## Code changes
- `Gemfile`, `Gemfile.lock`, `deqm_test_kit.gemspec` - I was having trouble with `bundle install` with sidekiq, so I added it as a dependency
- `lib/deqm_test_kit.rb` - Added Universal Realm STU 1 (https://hl7.org/fhir/uv/deqm/2026May/en/) Test Suite
- `lib/deqm-test-kit/collect_data_v1.rb` - test group for $collect-data operation tests for the Universal Realm STU 1 suite. Currently includes 4 success tests.
- `lib/deqm_test_kit/v1.0.0/deqm_test_suite.rb` - Universal Realm STU 1 Test Suite
- `measureUrlRadioButton.json` - Radio buttons that contain the `measureUrl` as the value
- `spec/deqm_test_kit/v1.0.0/collect_data_spec.rb` - simple spec tests collect-data test group.

# Testing guidance
- `bundle exec rubocop`
- `bundle exec rspec`
- Spin up this test kit with `ASYNC_JOBS=false bundle exec puma` (there are other ways to do this, but this makes it easy since it won't spin up the deqm-test-server and you won't have port conflicts)
- Spin up `bulk-export-server` with `npm run start` (this server contains the $collect-data endpoint that is best for testing this test group at the moment)
- `bulk-export-server` should be loaded with ecqm-content-qicore-2025 bundles. Note that bulk-export-server's $collect-data will have some valueset issues for certain Measures. Known measures that will result in passing tests are: CMS1017FHIRHHFI, CMS1056CTClinicalFHIR, CMS1074FHIRCTIQR, CMS117FHIRChildhoodImmunizationStatus, CMS1218FHIRHHRF, CMS122FHIRDiabetesAssessGreaterThan9Percent, and CMS124FHIRCervicalCancerScreening.
- Test the Universal Realm STU 1 test suite with url `http://localhost:3000` (or wherever your bulk-export-server is running)

### DIFF
--- a/lib/deqm_test_kit/collect_data_v1.rb
+++ b/lib/deqm_test_kit/collect_data_v1.rb
@@ -42,7 +42,16 @@ module DEQMTestKit
 
         parameters.parameter.each do |param|
           assert param.resource.is_a?(FHIR::Bundle), 'Expected parameter.resource to be a Bundle'
+          validate_bundles_contain_measure_report(param.resource)
         end
+      end
+
+      def validate_bundles_contain_measure_report(bundle)
+        assert bundle.entry.is_a?(Array), 'Expected Bundle.entry to be an array'
+        assert bundle.entry.any?, 'Expected at least one entry in the Bundle'
+
+        measure_reports = bundle.entry.map(&:resource).grep(FHIR::MeasureReport)
+        assert measure_reports.any?, 'Expected at least one MeasureReport in Bundle'
       end
 
       def collect_data_body(period_start:, period_end:, measure_urls:) # rubocop:disable Metrics/MethodLength
--- a/lib/deqm_test_kit/collect_data_v1.rb
+++ b/lib/deqm_test_kit/collect_data_v1.rb
@@ -18,9 +18,8 @@ module DEQMTestKit
         assert parameters.parameter.is_a?(Array), 'Expected Parameters.parameter to be an array'
         assert parameters.parameter.any?, 'Expected at least one parameter entry in Parameters resource'
 
-        parameters.parameter.each do |param|
-          assert param.resource.is_a?(FHIR::Bundle), 'Expected parameter.resource to be a Bundle'
-        end
+        # Check that first parameter is a bundle, checking all of them will cause the test to timeout
+        assert parameters.parameter[0].resource.is_a?(FHIR::Bundle), 'Expected parameter.resource to be a Bundle'
       end
 
       def collect_data_body(period_start:, period_end:, measure_urls:) # rubocop:disable Metrics/MethodLength
@@ -135,7 +134,8 @@ module DEQMTestKit
         assert result.resource.is_a?(FHIR::Parameters), "Expected
         resource to be a Parameters resource, but got #{result.resource&.class}"
 
-        # validate that the FHIR Parameters resource contains at least one FHIR Bundle
+        parameters = result.resource
+        validate_parameters_contains_bundles(parameters)
       end
     end
 
@@ -165,6 +165,9 @@ module DEQMTestKit
         assert_response_status(200)
         assert result.resource.is_a?(FHIR::Parameters), "Expected
         resource to be a Parameters resource, but got #{result.resource&.class}"
+
+        parameters = result.resource
+        validate_parameters_contains_bundles(parameters)
       end
     end
 
@@ -196,7 +199,8 @@ module DEQMTestKit
         assert result.resource.is_a?(FHIR::Parameters), "Expected
         resource to be a Parameters resource, but got #{result.resource&.class}"
 
-        # validate that the FHIR Parameters resource contains at least one FHIR Bundle
+        parameters = result.resource
+        validate_parameters_contains_bundles(parameters)
       end
     end
   end
--- a/lib/deqm_test_kit/collect_data_v1.rb
+++ b/lib/deqm_test_kit/collect_data_v1.rb
@@ -1,0 +1,204 @@
+# frozen_string_literal: true
+
+require 'json'
+
+module DEQMTestKit
+  # tests for $collect-data (DEQM UV v1.0.0)
+  # rubocop:disable Metrics/ClassLength
+  class CollectDataV1 < Inferno::TestGroup
+    # module for shared code for $collect-data assertions and requests
+    module CollectDataHelpers
+      def selected_measure_url(custom_url:, url:)
+        return custom_url.strip if url == 'Other' && custom_url&.strip&.length&.positive? # rubocop:disable Style/SafeNavigationChainLength
+
+        url
+      end
+
+      def validate_parameters_contains_bundles(parameters)
+        assert parameters.parameter.is_a?(Array), 'Expected Parameters.parameter to be an array'
+        assert parameters.parameter.any?, 'Expected at least one parameter entry in Parameters resource'
+
+        parameters.parameter.each do |param|
+          assert param.resource.is_a?(FHIR::Bundle), 'Expected parameter.resource to be a Bundle'
+        end
+      end
+
+      def collect_data_body(period_start:, period_end:, measure_urls:) # rubocop:disable Metrics/MethodLength
+        {
+          resourceType: 'Parameters',
+          parameter: [
+            *measure_urls.map do |url|
+              {
+                name: 'measureUrl',
+                valueCanonical: url
+              }
+            end,
+            {
+              name: 'periodStart',
+              valueDate: period_start
+            },
+            {
+              name: 'periodEnd',
+              valueDate: period_end
+            }
+          ]
+        }
+      end
+    end
+
+    id :collect_data_v1
+    title '$collect-data'
+    description 'Ensure FHIR server can perform the $collect-data operation'
+
+    fhir_client do
+      url :url
+      headers origin: url.to_s,
+              referrer: url.to_s,
+              'Content-Type': 'application/fhir+json'
+    end
+
+    measure_options = JSON.parse(File.read('./lib/fixtures/measureUrlRadioButton.json'))
+    measure_url_args = {
+      type: 'radio',
+      optional: false,
+      default: 'https://madie.cms.gov/Measure/CMS0334FHIRPCCesareanBirth',
+      options: measure_options,
+      title: 'Measure URL'
+    }
+    additional_measure_args = {
+      type: 'radio',
+      optional: false,
+      options: measure_options,
+      title: 'Measure URL for additional Measure'
+    }
+    custom_measure_url_args = {
+      type: 'text',
+      optional: true,
+      title: 'Custom Measure URL',
+      description: 'If you selected "Other" above, enter it here.'
+    }
+    custom_additional_measure_url_args = {
+      type: 'text',
+      optional: true,
+      title: 'Custom Additional Measure URL',
+      description: 'If you selected "Other" for the additional Measure URL, enter it here.'
+    }
+
+    test do
+      include CollectDataHelpers
+
+      title 'GET Measure/$collect-data with one measureUrl, periodStart, periodEnd'
+      id 'collect-data-one-measure-get'
+      description %(GET Measure/$collect-data with one measureUrl, periodStart, periodEnd returns 200 and
+      FHIR Parameters resource that contains at least one FHIR Bundle.)
+
+      input :measure_url, **measure_url_args
+      input :custom_measure_url, **custom_measure_url_args
+      input :period_start, title: 'Measurement Period Start', default: '2026-01-01'
+      input :period_end, title: 'Measurement Period End', default: '2026-12-31'
+
+      run do
+        body = collect_data_body(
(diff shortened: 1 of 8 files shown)


lib/deqm_test_kit/collect_data_v1.rb
```diff
@@ -0,0 +1,204 @@
...
+          measure_urls: [selected_measure_url(custom_url: custom_measure_url,
+                                              url: measure_url)], period_start: period_start, period_end: period_end
+        )
+
+        result = fhir_operation('/Measure/$collect-data', body: body)
+        assert_response_status(200)
+        assert result.resource.is_a?(FHIR::Parameters), "Expected
+        resource to be a Parameters resource, but got #{result.resource&.class}"
+
+        # validate that the FHIR Parameters resource contains at least one FHIR Bundle
```
dehall: I think you meant to call `validate_parameters_contains_bundles` here? Also line 199

elsaperelli: Yes I did 😄 added!

lib/deqm_test_kit/collect_data_v1.rb
```diff
@@ -0,0 +1,208 @@
...
+      input :period_start, title: 'Measurement Period Start', default: '2026-01-01'
+      input :period_end, title: 'Measurement Period End', default: '2026-12-31'
+
+      run do
+        body = collect_data_body(
+          measure_urls: [selected_measure_url(custom_url: custom_measure_url,
+                                              url: measure_url)], period_start: period_start, period_end: period_end
+        )
+
+        result = fhir_operation('/Measure/$collect-data', operation_method: :get,
```
lmd59, line 138: We shouldn't be sending a GET with a FHIR Parameters body (it should be url parameters), but it seems this is just an inferno weirdness where you can form the url parameters this way?

lib/deqm_test_kit/collect_data_v1.rb
```diff
@@ -0,0 +1,208 @@
...
+                                                          body: FHIR::Parameters.new(body))
+        assert_response_status(200)
+        assert result.resource.is_a?(FHIR::Parameters), "Expected
+        resource to be a Parameters resource, but got #{result.resource&.class}"
```
lmd59, line 142:
```suggestion
        resource to be a Parameters resource, but got #{result.resource&.class}"
        
        parameters = result.resource
        validate_parameters_contains_bundles(parameters)
```
Any reason not to include this for this test as well?

lib/deqm_test_kit/collect_data_v1.rb
```diff
@@ -0,0 +1,208 @@
...
+require 'json'
+
+module DEQMTestKit
+  # tests for $collect-data (DEQM UV v1.0.0)
+  # rubocop:disable Metrics/ClassLength
+  class CollectDataV1 < Inferno::TestGroup
+    # module for shared code for $collect-data assertions and requests
+    module CollectDataHelpers
+      def selected_measure_url(custom_url:, url:)
+        return custom_url.strip if url == 'Other' && custom_url&.strip&.length&.positive? # rubocop:disable Style/SafeNavigationChainLength
```
lmd59: If the user selects Other, should we disallow the custom url being empty, either through validation or some other error state?

lmd59 (commented):
Looks good! Some minor comments/questions inline. More generally:

- Recommend update PR description to outline expected data in bulk-export-server for test. Assuming ecqm-content-qicore-2025, it might be helpful to specify working measures that will have required valuesets without extra futzing. 124 and 1218 seemed to work well.
- Tests take a while to run due to running for all patients. Do we want all of them to be for all patients, or is it worth just doing one for all patients and doing the rest for a single subject or two (if the latter, this is something we can change later once the task for subject parameter comes up). (Also fine with leaving as is if you think it's critical that each of these test cases be tested with all available patients.)
- We validate that the returned Parameters object contains a bundle, but we should maybe consider a little more validation if it's not too strenuous. The description of return is:

> These Bundles SHALL contain 1..* DEQM Data Exchange MeasureReports with data-of-interest and SHOULD be for a single subject. See the [guidance on Bundle structure](https://hl7.org/fhir/uv/deqm/2026May/en/guidance.html#deqm-operation-bundles-organized-by-subject) for discussion about the Bundle content and organization.

We could validate that a data exchange MeasureReport is contained in the bundle for each of the measures selected.

lib/deqm_test_kit/collect_data_v1.rb
```diff
@@ -0,0 +1,208 @@
...
+
+        url
+      end
+
+      def validate_parameters_contains_bundles(parameters)
+        assert parameters.parameter.is_a?(Array), 'Expected Parameters.parameter to be an array'
+        assert parameters.parameter.any?, 'Expected at least one parameter entry in Parameters resource'
+
+        # Check that first parameter is a bundle, checking all of them will cause the test to timeout
```
dehall: I don't have the ability to run this yet (still waiting on docker access) but seeing this comment about a timeout is surprising. How many parameters are there expected to be? Seems like it should be pretty quick if it's just looping over an array and checking `is_a?`

spec/deqm_test_kit/v1.0.0/collect_data_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+    let(:period_end) { '2019-12-31' }
+
+    it 'passes with correct FHIR Parameters resource returned' do
+      parameters_response = create_parameters_response(measure_urls: [measure_url])
+
+      stub_request(
+        :post,
+        "#{url}/Measure/$collect-data"
+      ).with(
+        body: '{"resourceType":"Parameters","parameter":[{"name":"measureUrl","valueCanonical":"http://example.com/Measure/measure-EXM130"},{"name":"periodStart","valueDate":"2019-01-01"},{"name":"periodEnd","valueDate":"2019-12-31"}]}', # rubocop:disable Layout/LineLength
```
dehall: Minor but it would be nicer to read this if it weren't all on one line. Thoughts on a multiline string or building it as on object and calling `.to_json` ? Same in the other test on line 148 below

lib/deqm_test_kit/collect_data_v1.rb
elsaperelli: Since right now I was testing with bulk-export-server's implementation of $collect-data, I realize that my bulk-export-server currently has 2,225 patients (there should be a parameter for every patient so 2225 parameters). I can maybe add this check back then to check every Parameter since I think we normally expect fewer patients on the server under test. What do you think?

lib/deqm_test_kit/collect_data_v1.rb
elsaperelli: Yeah it is weird but confirmed it is sending it as url parameters!

image

spec/deqm_test_kit/v1.0.0/collect_data_spec.rb
elsaperelli: Good call! Addressed in [Add missing validation, make request body easier to read in spec test](https://github.com/projecttacoma/deqm-test-kit/pull/64/commits/867a4d33f3fc25083af841aee54508b2562086ef)

lib/deqm_test_kit/collect_data_v1.rb
elsaperelli: I forgot! Addressed in [Add missing validation, make request body easier to read in spec test](https://github.com/projecttacoma/deqm-test-kit/pull/64/commits/867a4d33f3fc25083af841aee54508b2562086ef)

lib/deqm_test_kit/collect_data_v1.rb
elsaperelli: Yes I agree! Unfortunately with the Inferno DSL, this isn't super easy and seems to require a good chunk of additional functionality. What are your thoughts on me creating a backlog task for this? I don't think it's super high priority compared to overall test kit functionality, and this was the behavior in the STU 5 test suite.

elsaperelli: Actually I did add some runtime checks for this! [Add runtime checks for radio buttons and other](https://github.com/projecttacoma/deqm-test-kit/pull/64/commits/dc97efd11e5a269f4842e706680ab73de67bb46c)

lib/deqm_test_kit/collect_data_v1.rb
dehall: Hm I guess I can see why a test involving 2,225 patients would timeout, but I don't really see why this line would be what causes it. Is there maybe some lazy-loading that doesn't parse json to a FHIR object until it's called or something? I do think in general it's better to check everything rather than just the first item, but also it's not worth spending a ton of time trying to figure out why this makes it slow unless it's something we think an implementer might get wrong

elsaperelli: I just tested it again and it was totally fine so I think I missed something or got confused. Added that check back, sorry about that!

elsaperelli:
> Looks good! Some minor comments/questions inline. More generally:
> 
> * Recommend update PR description to outline expected data in bulk-export-server for test. Assuming ecqm-content-qicore-2025, it might be helpful to specify working measures that will have required valuesets without extra futzing. 124 and 1218 seemed to work well.
> * Tests take a while to run due to running for all patients. Do we want all of them to be for all patients, or is it worth just doing one for all patients and doing the rest for a single subject or two (if the latter, this is something we can change later once the task for subject parameter comes up). (Also fine with leaving as is if you think it's critical that each of these test cases be tested with all available patients.)
> * We validate that the returned Parameters object contains a bundle, but we should maybe consider a little more validation if it's not too strenuous. The description of return is:
> 
> > These Bundles SHALL contain 1..* DEQM Data Exchange MeasureReports with data-of-interest and SHOULD be for a single subject. See the [guidance on Bundle structure](https://hl7.org/fhir/uv/deqm/2026May/en/guidance.html#deqm-operation-bundles-organized-by-subject) for discussion about the Bundle content and organization.
> 
> We could validate that a data exchange MeasureReport is contained in the bundle for each of the measures selected.

1. Good call. Added that to the PR description!
2. This is fair, although I don't think they take tooo long. Thoughts on leaving them in for now since they are done and then we can evaluate the structure later when everything is done?
3. I initially wrote the task to just check for Bundles for now, but I am realizing now that this is an easy add anyways so added! Good suggestion.

lib/deqm_test_kit/collect_data_v1.rb
```diff
@@ -42,9 +42,18 @@ def validate_parameters_contains_bundles(parameters)
...
+          validate_bundles_contain_measure_report(param.resource)
         end
       end
 
+      def validate_bundles_contain_measure_report(bundle)
+        assert bundle.entry.is_a?(Array), 'Expected Bundle.entry to be an array'
+        assert bundle.entry.any?, 'Expected at least one entry in the Bundle'
+
+        measure_reports = bundle.entry.map(&:resource).grep(FHIR::MeasureReport)
+        assert measure_reports.any?, 'Expected at least one MeasureReport in Bundle'
```
lmd59: This is not too important, but it seems like it would be pretty straightforward to pass a number into this function based on how many measures we're testing against and actually check for exactly that number of measure reports? Happy to ignore if it's annoying for some reason I don't see though.
Could even pass the exact measures selected through and check that the correct MeasureReport exists (MeasureReport.measure should be the canonical).

lmd59 (commented):
Thank you for updates! Agreed to your responses on general updates!

lib/deqm_test_kit/collect_data_v1.rb
elsaperelli: Good idea! I just added a check for the number of MeasureReports for now. Addressed in [Add check for number of MeasureReports](https://github.com/projecttacoma/deqm-test-kit/pull/64/commits/c5bc1ccc9ca95fa31494fdd2f86e0f4bb2109bb1)

lmd59 (approved):
Looking good!

PR #65, opened 2026-07-06
Collect data fail tests

ptran265:
# Summary
This PR adds 6 fail tests for the $collect-data operation in the Universal Realm STU 1 Test Suite.
## New behavior
The 6 $collect-data failure tests are as follows:
- GET Measure/$collect-data with measureUrl and periodStart in request parameters returns 400 and FHIR Operation Outcome when periodEnd was missing.
- POST Measure/$collect-data with measureUrl and periodStart in request body returns 400 and FHIR Operation Outcome when periodEnd was missing.
- GET Measure/$collect-data with measureUrl and periodEnd in request parameters returns 400 and FHIR Operation Outcome when periodStart was missing.
- POST Measure/$collect-data with measureUrl and periodEnd in request body returns 400 and FHIR Operation Outcome when periodStart was missing.
- GET Measure/$collect-data with periodStart and periodEnd in request parameters returns 400 and FHIR Operation Outcome when measureUrl was missing.
- POST Measure/$collect-data with periodStart and periodEnd in request body returns 400 and FHIR Operation Outcome when measureUrl was missing.
## Code changes
lib/deqm-test-kit/collect_data_v1.rb - Added 6 failure tests.
# Testing guidance
- `bundle exec rubocop`
- `bundle exec rspec`
- Spin up this test kit with `ASYNC_JOBS=false bundle exec puma` 
- Spin up `bulk-export-server` with `npm run start` 
- `bulk-export-server` should be loaded with ecqm-content-qicore-2025 bundles. Note that bulk-export-server's $collect-data will have some valueset issues for certain Measures. Known measures that will result in passing tests are: CMS1017FHIRHHFI, CMS1056CTClinicalFHIR, CMS1074FHIRCTIQR, CMS117FHIRChildhoodImmunizationStatus, CMS1218FHIRHHRF, CMS122FHIRDiabetesAssessGreaterThan9Percent, and CMS124FHIRCervicalCancerScreening.
- Test the Universal Realm STU 1 test suite with url `http://localhost:3000` (or wherever your bulk-export-server is running)

### DIFF
--- a/lib/deqm_test_kit/collect_data_v1.rb
+++ b/lib/deqm_test_kit/collect_data_v1.rb
@@ -239,6 +239,39 @@ module DEQMTestKit
         validate_parameters_contains_bundles(parameters, 2)
       end
     end
+
+    # GET Measure/$collect-data with measureUrl and periodStart in request parameters returns 400 and
+    # FHIR Operation Outcome when periodEnd was missing.
+    test do
+      title 'GET Measure/$collect-data with one measureUrl, periodStart'
+      id 'collect-data-missing-period-end-fail'
+      description %(GET Measure/$collect-data with one measureUrl and periodStart returns 400 and
+      FHIR OperationOutcome when periodEnd was missing.)
+
+      input :measure_url, **measure_url_args
+      input :custom_measure_url, **custom_measure_url_args
+      input :period_start, title: 'Measurement Period Start', default: '2026-01-01'
+
+      run do
+        body = {
+          resourceType: 'Parameters',
+          parameter: [
+            {
+              name: 'measureUrl',
+              valueCanonical: measure_url
+            },
+            {
+              name: 'periodStart',
+              valueDate: period_start
+            }
+          ]
+        }
+
+        result = fhir_operation('/Measure/$collect-data', operation_method: :get, body: FHIR::Parameters.new(body))
+        assert_response_status(400)
+        assert result.resource.is_a?(FHIR::OperationOutcome)
+      end
+    end
   end
   # rubocop:enable Metrics/ClassLength
 end
--- a/lib/deqm_test_kit/collect_data_v1.rb
+++ b/lib/deqm_test_kit/collect_data_v1.rb
@@ -243,8 +243,10 @@ module DEQMTestKit
     # GET Measure/$collect-data with measureUrl and periodStart in request parameters returns 400 and
     # FHIR Operation Outcome when periodEnd was missing.
     test do
+      include CollectDataHelpers
+
       title 'GET Measure/$collect-data with one measureUrl, periodStart'
-      id 'collect-data-missing-period-end-fail'
+      id 'collect-data-missing-period-end-get-fail'
       description %(GET Measure/$collect-data with one measureUrl and periodStart returns 400 and
       FHIR OperationOutcome when periodEnd was missing.)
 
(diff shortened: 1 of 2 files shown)


lib/deqm_test_kit/collect_data_v1.rb
```diff
@@ -239,6 +239,190 @@ def collect_data_body(period_start:, period_end:, measure_urls:) # rubocop:disab
         validate_parameters_contains_bundles(parameters, 2)
       end
     end
+
+    # GET Measure/$collect-data with measureUrl and periodStart in request parameters returns 400 and
+    # FHIR Operation Outcome when periodEnd was missing.
+    test do
+      include CollectDataHelpers
+
+      title 'GET Measure/$collect-data with one measureUrl, periodStart'
```
dehall: This follows the convention previously set in the file but ideally test titles should mention the expectation of what happens, and for an error test, highlight what the problem in the scenario is. For example here I would suggest:
```suggestion
      title 'GET Measure/$collect-data missing periodEnd returns HTTP 400 and OperationOutcome'
```

(feel free to wordsmith further)

lib/deqm_test_kit/collect_data_v1.rb
```diff
@@ -239,6 +239,190 @@ def collect_data_body(period_start:, period_end:, measure_urls:) # rubocop:disab
...
+
+      run do
+        selected_url = selected_measure_url(custom_url: custom_measure_url, url: measure_url)
+        body = {
+          resourceType: 'Parameters',
+          parameter: [
+            { name: 'measureUrl', valueCanonical: selected_url },
+            { name: 'periodStart', valueDate: period_start }
+          ]
+        }
```
dehall: For discussion, not necessarily requesting a change:
This is totally fine, it's explicit what the request body is, but as with the title where it's nice to highlight what's missing, I'd propose for consideration starting with a good request and deleting a missing piece.

eg (I didn't test this but it should give you the idea)

```ruby
body = collect_data_body(...) # as called in above test
body.parameters.delete_if(p => p.name == 'periodEnd')
```

dehall (commented):
This is missing spec tests (see spec/deqm_test_kit/v1.0.0/collect_data_spec.rb) but I think you should be able to copy & paste the existing ones to create ones for these pretty easily.
(Personally I feel like spec tests for tests like these are unnecessary but the convention is that they exist)

Also: do we want other failure cases, like invalid date formats, etc? I realize this was the set listed in the ticket so no need to implement more here, but we could add another ticket if we want.

Other than that it looks good, but I do have a couple comments primarily around conventions in this file, so I'd defer to @elsaperelli or @lmd59 on those.

elsaperelli:
> This is missing spec tests (see spec/deqm_test_kit/v1.0.0/collect_data_spec.rb) but I think you should be able to copy & paste the existing ones to create ones for these pretty easily. (Personally I feel like spec tests for tests like these are unnecessary but the convention is that they exist)
> 
> Also: do we want other failure cases, like invalid date formats, etc? I realize this was the set listed in the ticket so no need to implement more here, but we could add another ticket if we want.
> 
> Other than that it looks good, but I do have a couple comments primarily around conventions in this file, so I'd defer to @elsaperelli or @lmd59 on those.

Yes spec tests are nice to have if they are easy but also fine to skip them for now while we prepare for the Connectathon and I can always add tasks for adding them in later.

I will add more failure cases in tickets in the backlog!

I will take a look at the individual comments.

lib/deqm_test_kit/collect_data_v1.rb
elsaperelli: This is a great suggestion. I am fine with Patricia leaving it as is (I kinda pointed her to this format based on other tests in the suite so I feel bad having her change it). But I will make a backlog task for this. Unless @ptran265 you want to try making this update!

lib/deqm_test_kit/collect_data_v1.rb
elsaperelli: I agree. I would format all of the titles and descriptions around Dylan's suggestion here.

lib/deqm_test_kit/collect_data_v1.rb
ptran265: If I do like that then won't I need to change the inputs as well, in order to start with a good request? 
`input :period_end, title: 'Measurement Period End', default: '2026-12-31'`

elsaperelli: That is true, I think you could either add the missing input or provide it with a dummy value like

```
body = collect_data_body(period_end: '2026-12-31', ...)
```

Since it will get removed anyways

ptran265: I tried making that update, but `body.parameters.delete_if(p => p.name == 'periodEnd')` did not work, so codex gave me this alternative `body[:parameter].delete_if { |param| param[:name] == 'periodEnd' }`. ~~However, I'm not sure if this is just me but this didn't seem to pass reliably - now it seems the first 4 original tests (the success tests) aren't passing while my fail tests are passing~~ Scratch that - I accidentally deleted all my patients in the server...

dehall: Oh yeah, ruby uses blocks not arrow syntax, it's been a while... As of the last edit it sounds like everything is ok, but in the future if a minor change for style reasons like this breaks something, I'd say just skip the change. Another approach we could have gone with here is just call out with a comment that the field was excluded, eg:

```ruby
body = {
          resourceType: 'Parameters',
          parameter: [
            { name: 'measureUrl', valueCanonical: selected_url },
            { name: 'periodStart', valueDate: period_start }
            # periodEnd intentionally omitted
          ]
        }
```

dehall: I hate to say this but now that I actually see the approach of adding a dummy value just to delete it, I don't love it. I won't hold up the PR just for that since it works and again it's a style thing, but if you do prefer the other approach I'll leave it up to you which one you want

dehall (dismissed):

lib/deqm_test_kit/collect_data_v1.rb
ptran265: Hm I think I like the old style with the comment like you suggested; it's feels easier to read to me. I reverted it since it was an easy change. I also added the spec tests.

dehall (approved):

PR #67, opened 2026-07-09
Stu collect data subject

elsaperelli:
# Summary
This PR adds tests for $collect-data requests with the `subject` parameter included with a Patient reference.

## New behavior
Four new tests added to the DEQM UV STU1 Test Suite for the $collect-data test suite.

## Code changes
- `lib/deqm_test_kit/collect_data_v1.rb` - add method `validate_number_of_bundles` to check that the right number of bundles are returned, update `collect_data_body` method to include optional `subject` parameter for some tests, 4 new subject parameter tests
- `spec/deqm_test_kit/v1.0.0/collect_data_spec.rb` - new associated spec tests

# Testing guidance
In deqm-test-kit:
- `bundle exec rspec`
- `bundle exec rubocop`
- `ASYNC_JOBS=false bundle exec puma`
- Inputs to use:
    - Measure URL: CMS1017FHIRHHFI
    - url: http://localhost:3001
    - Measure URL for additional Measure: `CMS1047FHIRCTIQR`
    - Patient ID: f6dafa7e-15a9-4fec-baea-59d43201caf8

In bulk-export-server (at the same time):
- `npm run start`

### DIFF
--- a/lib/deqm_test_kit/collect_data_v1.rb
+++ b/lib/deqm_test_kit/collect_data_v1.rb
@@ -240,6 +240,143 @@ module DEQMTestKit
       end
     end
 
+    test do
+      include CollectDataHelpers
+
+      title 'GET Measure/$collect-data with one measureUrl, periodStart, periodEnd, and subject=Patient/patientId'
+      id 'collect-data-one-measure-get-subject-patient'
+      description %(GET Measure/$collect-data with one measureUrl, periodStart, periodEnd, and subject=Patient/patientId
+      returns 200 and FHIR Parameters resource that contains exactly one FHIR Bundle that contains one MeasureReport)
+
+      input :measure_url, **measure_url_args
+      input :custom_measure_url, **custom_measure_url_args
+      input :period_start, title: 'Measurement Period Start', default: '2026-01-01'
+      input :period_end, title: 'Measurement Period End', default: '2026-12-31'
+      input :patient_id, title: 'Patient ID'
+
+      run do
+        body = collect_data_body(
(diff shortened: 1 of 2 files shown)


lib/deqm_test_kit/collect_data_v1.rb
```diff
@@ -240,6 +254,149 @@ def collect_data_body(period_start:, period_end:, measure_urls:) # rubocop:disab
...
+                                                          body: FHIR::Parameters.new(body))
+
+        assert_response_status(200)
+        assert result.resource.is_a?(FHIR::Parameters), "Expected
+        resource to be a Parameters resource, but got #{result.resource&.class}"
+
+        parameters = result.resource
+
+        validate_number_of_bundles(parameters, 1)
+        validate_parameters_contains_bundles(parameters, 1)
```
ptran265, line 287: I'm not sure if this matters but do you think `validate_parameters_contains_bundles` should be called before `validate_number_of_bundles`? Because when I look at the assertions in each one, `validate_parameters_contains_bundles` checks that Parameters.parameter is present and is an array - which would make sense to check _before_ checking the length.

ptran265 (commented):
This looks good and passes as described. I just left one comment for your consideration

lib/deqm_test_kit/collect_data_v1.rb
elsaperelli: I think that's a good call!

ptran265 (approved):
Looks good to me! Thanks for adding the other 2 spec tests and for making that change with the `validate_number_of_bundles`

PR #69, opened 2026-07-14
collect-data with dataEndpoint param tests

elsaperelli:
# Summary
This PR adds two `$collect-data` operation tests, specifically for a POST to `$collect-data` with the `dataEndpoint` parameter.

## New behavior
New test group added (separated this from the other `$collect-data` test group because our implementations of `$collect-data` are different between our `bulk-export-server` (doesn't support `dataEndpoint`) and `deqm-test-server` (only supports `dataEndoint`). Also adds some STU1 test suite presets.

## Code changes
- `config/presents/` - added presets for local and flame-demo instances of `deqm-test-server` and `bulk-export-server`. `bulk-export-server` used for the $collect-data (without dataEndpoint) test group, `deqm-test-server` used for the rest
- `collect_data_endpoint_v1.rb` - collect-data test group separate from collect-data because it requires the dataEndpoint parameter
- `measureUrlRadioButton.json` - updated radio buttons to match dqm-content-qicore-2025.

# Testing guidance
In deqm-test-kit:
- `bundle exec rubocop`
- `bundle exec rspec` (didn't add any spec tests yet to focus on the cthon)
- `ASYNC_JOBS=false bundle exec puma`
- Use the DEQM Test Server Local preset

In deqm-test-server:
- `npm run start`

### DIFF
--- a/lib/fixtures/measureUrlRadioButton.json
+++ b/lib/fixtures/measureUrlRadioButton.json
@@ -1,44 +1,56 @@
 {
   "list_options": [
     {
-      "label": "CMS0334FHIRPCCesareanBirth",
-      "value": "https://madie.cms.gov/Measure/CMS0334FHIRPCCesareanBirth"
-    },
-    {
-      "label": "CMS1017FHIRHHFI",
-      "value": "https://madie.cms.gov/Measure/CMS1017FHIRHHFI"
-    },
-    {
-      "label": "CMS1056CTClinicalFHIR",
-      "value": "https://madie.cms.gov/Measure/CMS1056CTClinicalFHIR"
-    },
-    {
-      "label": "CMS1074FHIRCTIQR",
-      "value": "https://madie.cms.gov/Measure/CMS1074FHIRCTIQR"
-    },
-    {
-      "label": "CMS108FHIRVTEProphylaxis",
-      "value": "https://madie.cms.gov/Measure/CMS108FHIRVTEProphylaxis"
-    },
-    {
-      "label": "CMS117FHIRChildhoodImmunizationStatus",
-      "value": "https://madie.cms.gov/Measure/CMS117FHIRChildhoodImmunizationStatus"
-    },
-    {
-      "label": "CMS1218FHIRHHRF",
-      "value": "https://madie.cms.gov/Measure/CMS1218FHIRHHRF"
+      "label": "CMS2FHIRPCSDepScreenAndFollowUp",
+      "value": "https://madie.cms.gov/Measure/CMS2FHIRPCSDepScreenAndFollowUp"
     },
     {
       "label": "CMS122FHIRDiabetesAssessGreaterThan9Percent",
       "value": "https://madie.cms.gov/Measure/CMS122FHIRDiabetesAssessGreaterThan9Percent"
     },
     {
-      "label": "CMS1244FHIRECATHOQR",
-      "value": "https://madie.cms.gov/Measure/CMS1244FHIRECATHOQR"
-    },
-    {
       "label": "CMS124FHIRCervicalCancerScreening",
       "value": "https://madie.cms.gov/Measure/CMS124FHIRCervicalCancerScreening"
+    },
+    {
+      "label": "CMS125FHIRBreastCancerScreen",
+      "value": "https://madie.cms.gov/Measure/CMS125FHIRBreastCancerScreen"
+    },
+    {
+      "label": "CMS130FHIRColorectalCancerScrn",
+      "value": "https://madie.cms.gov/Measure/CMS130FHIRColorectalCancerScrn"
+    },
+    {
+      "label": "CMS165FHIRControllingHighBP",
+      "value": "https://madie.cms.gov/Measure/CMS165FHIRControllingHighBP"
+    },
+    {
+      "label": "CMSFHIR529HybridHospitalWideReadmission",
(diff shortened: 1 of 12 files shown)


lib/fixtures/measureUrlRadioButton.json
```diff
@@ -1,44 +1,56 @@
...
-      "label": "CMS0334FHIRPCCesareanBirth",
-      "value": "https://madie.cms.gov/Measure/CMS0334FHIRPCCesareanBirth"
+      "label": "CMS2FHIRPCSDepScreenAndFollowUp",
+      "value": "https://madie.cms.gov/Measure/CMS2FHIRPCSDepScreenAndFollowUp"
     },
     {
-      "label": "CMS1017FHIRHHFI",
-      "value": "https://madie.cms.gov/Measure/CMS1017FHIRHHFI"
+      "label": "CMS122FHIRDiabetesAssessGreaterThan9Percent",
+      "value": "https://madie.cms.gov/Measure/CMS122FHIRDiabetesAssessGreaterThan9Percent"
```
lmd59:
```suggestion
      "label": "CMS122FHIRDiabetesAssessGT9Pct",
      "value": "https://madie.cms.gov/Measure/CMS122FHIRDiabetesAssessGT9Pct"
```

lib/fixtures/measureUrlRadioButton.json
```diff
@@ -1,44 +1,56 @@
...
     },
     {
-      "label": "CMS1056CTClinicalFHIR",
-      "value": "https://madie.cms.gov/Measure/CMS1056CTClinicalFHIR"
+      "label": "CMS124FHIRCervicalCancerScreening",
+      "value": "https://madie.cms.gov/Measure/CMS124FHIRCervicalCancerScreening"
```
lmd59:
```suggestion
      "label": "CMS124FHIRCervicalCancerScreen",
      "value": "https://madie.cms.gov/Measure/CMS124FHIRCervicalCancerScreen"
```

lib/fixtures/measureUrlRadioButton.json
lmd59: Since we're updating according to correspond with dqm-content-qicore-2025, I believe this is the expected URL

lib/fixtures/measureUrlRadioButton.json
lmd59: Since we're updating according to correspond with dqm-content-qicore-2025, I believe this is the expected URL

lmd59 (commented):
Works well! The exception I saw was some discrepancies in urls for dqm-content-qicore-2025.
For code quality moving forward...
It definitely seems like there's some code repeated here for expediency during the connectathon. Can we centralize that into a reusable utility across test sets?
Additionally, now that connectathon is wrapping up, maybe we can add those spec tests?

spec/deqm_test_kit/v1.0.0/collect_data_endpoint_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+RSpec.describe DEQMTestKit::CollectDataEndpointV1 do
+  include DEQMTestKit::CollectDataSpecUtils
+
+  let(:suite) { Inferno::Repositories::TestSuites.new.find('deqm_v100') }
+  let(:group) { suite.groups[3] }
+  let(:session_data_repo) { Inferno::Repositories::SessionData.new }
+  let(:test_session) { repo_create(:test_session, test_suite_id: suite.id) }
+  let(:url) { 'http://example.com/fhir' }
+  let(:data_endpoint) do
+    {
```
lmd59, line 14: Just to make this valid, I would add in a payload type since it is required. Something like:
```
"payloadType": [
					{
						"coding": [
							{
								"system": "http://hl7.org/fhir/resource-types",
								"code": "Parameters"
							}
						]
					}
				],
```

spec/deqm_test_kit/v1.0.0/collect_data_endpoint_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      resourceType: 'Endpoint',
+      status: 'active',
+      connectionType: { system: 'http://terminology.hl7.org/CodeSystem/endpoint-connection-type',
+                        code: 'hl7-fhir-rest' },
+      address: 'https://r4.smarthealthit.org'
```
lmd59: Maybe make this an example.com address or something a bit more generic since this is a mock endpoint.

lmd59 (commented):
Two small inline comments, and then one more suggestion that I think will be a pretty straightforward add (let me know if it doesn't make sense at all!):
It looks like we haven't done much of these in the 1.0.0 suite yet, but it would be nice to add a couple of failure spec tests to get the ball rolling on those. Maybe add a 'fails if result if not a Parameters resource' and 'fails if result has incorrect number of measure reports'? 

I would just add both to the `describe 'POST Measure/$collect-data with two measureUrls, subject, and dataEndpoint'` spec test group.
For the first one, you can just create a bad dummy parameters response (i.e. something like `{resourceType: 'Invalid'}`). For the second one, just create the parameters response with one measure url (i.e. `create_parameters_response(measure_urls: [measure_url])), which will result in the wrong number of measure reports.

ptran265:
@lmd59 I added the two failure spec tests and made the other two changes you pointed out. Thanks for the review! Let me know if there's anything else I should fix.

lmd59 (approved):
Awesome! Looks good to me!

PR #70, opened 2026-07-29
add collect_data tests with subjectGroup parameter

ptran265:
# Summary
This PR adds two $collect-data operation tests, specifically for a POST to $collect-data with the subjectGroup parameter.
## New behavior
- POST Measure/$collect-data with 1 measureUrl, periodStart, periodEnd, subjectGroup in the request body returns 200 with FHIR Parameters resource with X number of Bundles (how many patients are in the subjectGroup passed in) each with 1 Measure Report.
- POST Measure/$colelct-data with 2 measureUrls, periodStart, periodEnd, subjectGroup in the request body returns 200 with FHIR Parameters resource with X number of Bundles (how many patients are in the subjectGroup passed in) each with 2 Measure Reports.

## Code changes
`lib/deqm-test-kit/collect_data_v1.rb` - Added 2 tests as described above.
`lib/utils/collect_data_utils.rb` - updated `collect_data_body` to take a list of patients (`patient_id_list`)
`spec/deqm_test_kit/v1.0.0/collect_data_spec.rb` - Added corresponding spec pass and fail tests.
`spec/utils/collect_data_spec_utils.rb` - updated `create_parameters_request` to take a list of patients (`patient_id_list`) 

# Testing guidance
In deqm-test-kit:
`bundle exec rubocop`
`bundle exec rspec `
`ASYNC_JOBS=false bundle exec puma`

In bulk-export-server:
`npm run start`

For the inputs, select CMS506FHIRSafeUseofOpioids, CMS1017FHIRHHFI as the measures, http://localhost:3001 as the url, enter f6dafa7e-15a9-4fec-baea-59d43201caf8 as Patient ID, and f6dafa7e-15a9-4fec-baea-59d43201caf8, c58acff5-248b-49c9-b18d-69e4a84a08d9 in the Patient IDs list

### DIFF
--- a/lib/deqm_test_kit/collect_data_v1.rb
+++ b/lib/deqm_test_kit/collect_data_v1.rb
@@ -312,6 +312,88 @@ module DEQMTestKit
       end
     end
 
+    test do # rubocop:disable Metrics/BlockLength
+      include CollectDataUtils
+
+      title 'POST Measure/$collect-data with one measureUrl, periodStart, periodEnd, and subjectGroup'
+      id 'collect-data-one-measure-post-subject-group'
+      description %(POST Measure/$collect-data with one measureUrl, periodStart, periodEnd, and
+      subjectGroup returns 200 and FHIR Parameters resource that contains X number of FHIR Bundles that
+      each contain one MeasureReport)
+
+      input :measure_url, **measure_url_args
+      input :custom_measure_url, **custom_measure_url_args
+      input :period_start, title: 'Measurement Period Start', default: '2026-01-01'
+      input :period_end, title: 'Measurement Period End', default: '2026-12-31'
+      input :patient_ids,
+            title: 'Patient IDs',
+            description: 'Enter a comma-delimited list of patient IDs.'
+
+      run do
+        patient_id_list = patient_ids.split(',').map(&:strip).reject(&:empty?)
+
+        body = collect_data_body(
+          measure_urls: [selected_measure_url(custom_url: custom_measure_url, url: measure_url)],
+          period_start: period_start,
+          period_end: period_end,
+          patient_id_list: patient_id_list
+        )
+
+        result = fhir_operation('/Measure/$collect-data', body: body)
+
+        assert_response_status(200)
+        assert result.resource.is_a?(FHIR::Parameters), "Expected
+        resource to be a Parameters resource, but got #{result.resource&.class}"
+
+        parameters = result.resource
+
+        validate_parameters_contains_bundles(parameters, 1)
+        validate_number_of_bundles(parameters, patient_id_list.length)
+      end
+    end
+
+    test do # rubocop:disable Metrics/BlockLength
+      include CollectDataUtils
+
+      title 'POST Measure/$collect-data with two measureUrls, periodStart, periodEnd, and subjectGroup'
+      id 'collect-data-two-measures-post-subject-group'
+      description %(POST Measure/$collect-data with two measureUrls, periodStart, periodEnd, and
+      subjectGroup returns 200 and FHIR Parameters resource that contains X number of FHIR Bundles that
+      each contain two MeasureReports)
+
+      input :measure_url, **measure_url_args
+      input :custom_measure_url, **custom_measure_url_args
+      input :additional_measure_url, **additional_measure_args
+      input :custom_additional_measure_url, **custom_additional_measure_url_args
+      input :period_start, title: 'Measurement Period Start', default: '2026-01-01'
+      input :period_end, title: 'Measurement Period End', default: '2026-12-31'
+      input :patient_ids,
+            title: 'Patient IDs',
+            description: 'Enter a comma-delimited list of patient IDs.'
+
+      run do
+        patient_id_list = patient_ids.split(',').map(&:strip).reject(&:empty?)
+
+        body = collect_data_body(
+          measure_urls: selected_measure_urls,
+          period_start: period_start,
+          period_end: period_end,
+          patient_id_list: patient_id_list
+        )
+
+        result = fhir_operation('/Measure/$collect-data', body: body)
+
+        assert_response_status(200)
+        assert result.resource.is_a?(FHIR::Parameters), "Expected
+        resource to be a Parameters resource, but got #{result.resource&.class}"
+
+        parameters = result.resource
+
+        validate_parameters_contains_bundles(parameters, 2)
+        validate_number_of_bundles(parameters, patient_id_list.length)
+      end
+    end
+
     # GET Measure/$collect-data with measureUrl and periodStart in request parameters returns 400 and
     # FHIR Operation Outcome when periodEnd was missing.
     test do
--- a/lib/deqm_test_kit/collect_data_v1.rb
+++ b/lib/deqm_test_kit/collect_data_v1.rb
@@ -318,8 +318,8 @@ module DEQMTestKit
       title 'POST Measure/$collect-data with one measureUrl, periodStart, periodEnd, and subjectGroup'
       id 'collect-data-one-measure-post-subject-group'
       description %(POST Measure/$collect-data with one measureUrl, periodStart, periodEnd, and
-      subjectGroup returns 200 and FHIR Parameters resource that contains X number of FHIR Bundles that
-      each contain one MeasureReport)
+      subjectGroup returns 200 and FHIR Parameters resource that contains X (patient count)
+      number of FHIR Bundles that each contain one MeasureReport)
 
       input :measure_url, **measure_url_args
       input :custom_measure_url, **custom_measure_url_args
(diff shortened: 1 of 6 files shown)


lib/deqm_test_kit/collect_data_v1.rb
```diff
@@ -312,6 +312,88 @@ class CollectDataV1 < Inferno::TestGroup
       end
     end
 
+    test do # rubocop:disable Metrics/BlockLength
+      include CollectDataUtils
+
+      title 'POST Measure/$collect-data with one measureUrl, periodStart, periodEnd, and subjectGroup'
+      id 'collect-data-one-measure-post-subject-group'
+      description %(POST Measure/$collect-data with one measureUrl, periodStart, periodEnd, and
+      subjectGroup returns 200 and FHIR Parameters resource that contains X number of FHIR Bundles that
```
lmd59:
```suggestion
      subjectGroup returns 200 and FHIR Parameters resource that contains X (patient count) number of FHIR Bundles that
```

lib/deqm_test_kit/collect_data_v1.rb
```diff
@@ -312,6 +312,88 @@ class CollectDataV1 < Inferno::TestGroup
...
+      end
+    end
+
+    test do # rubocop:disable Metrics/BlockLength
+      include CollectDataUtils
+
+      title 'POST Measure/$collect-data with two measureUrls, periodStart, periodEnd, and subjectGroup'
+      id 'collect-data-two-measures-post-subject-group'
+      description %(POST Measure/$collect-data with two measureUrls, periodStart, periodEnd, and
+      subjectGroup returns 200 and FHIR Parameters resource that contains X number of FHIR Bundles that
```
lmd59:
```suggestion
      subjectGroup returns 200 and FHIR Parameters resource that contains X (patient count) number of FHIR Bundles that
```

lib/utils/collect_data_utils.rb
```diff
@@ -72,6 +74,19 @@ def collect_data_body(period_start:, period_end:, measure_urls:, patient_id: nil
         }
       end
 
+      if patient_id_list
+        parameters << {
+          name: 'subjectGroup',
+          resource: {
```
lmd59, line 79: Group resource should include `type` and `actual` fields (required by https://hl7.org/fhir/R4/group.html to be valid).

lib/deqm_test_kit/collect_data_v1.rb
```diff
@@ -312,6 +312,88 @@ class CollectDataV1 < Inferno::TestGroup
...
+      id 'collect-data-one-measure-post-subject-group'
+      description %(POST Measure/$collect-data with one measureUrl, periodStart, periodEnd, and
+      subjectGroup returns 200 and FHIR Parameters resource that contains X number of FHIR Bundles that
+      each contain one MeasureReport)
+
+      input :measure_url, **measure_url_args
+      input :custom_measure_url, **custom_measure_url_args
+      input :period_start, title: 'Measurement Period Start', default: '2026-01-01'
+      input :period_end, title: 'Measurement Period End', default: '2026-12-31'
+      input :patient_ids,
```
lmd59, line 328: Since patient_ids is a newly added input, consider adding appropriate patient_ids input to the presets for v100 in config/presets

lib/utils/collect_data_utils.rb
```diff
@@ -57,7 +57,9 @@ def validate_bundles_contain_measure_report(bundle, measure_count)
              "Expected #{measure_count} MeasureReport(s), got #{measure_reports.length}"
     end
 
-    def collect_data_body(period_start:, period_end:, measure_urls:, patient_id: nil, data_endpoint: nil) # rubocop:disable Metrics/MethodLength
+    # rubocop:disable Metrics/MethodLength, Metrics/ParameterLists
+    def collect_data_body(period_start:, period_end:, measure_urls:, patient_id: nil,
```
lmd59: Change collect_data_body function signature to take in "options" (or similar) object instead which contains these 6 parameters. This changewould allow re-enabling rubocop ParameterLists check as well.

spec/utils/collect_data_spec_utils.rb
```diff
@@ -18,9 +18,21 @@ def create_parameters_response(measure_urls:)
       FHIR::Parameters.new(parameter: { name: 'return', resource: bundle })
     end
 
-    def create_parameters_request(measure_urls:, period_start:, period_end:, patient_id: nil, data_endpoint: nil)
+    # rubocop:disable Metrics/MethodLength, Metrics/ParameterLists
+    def create_parameters_request(measure_urls:, period_start:, period_end:, patient_id: nil, patient_id_list: nil,
```
lmd59: Similar recommendation to condense inputs into an options object and restore rubocop ParameterLists check.

spec/deqm_test_kit/v1.0.0/collect_data_spec.rb
```diff
@@ -263,6 +263,224 @@
...
+    let(:period_start) { '2019-01-01' }
+    let(:period_end) { '2019-12-31' }
+    let(:patient_ids) { 'patient-1, patient-2' }
+
+    it 'passes with correct FHIR Parameters resource returned' do
+      patient_id_list = patient_ids.split(',').map(&:strip).reject(&:empty?)
+      parameters_request = create_parameters_request(
+        measure_urls: [measure_url], period_start:, period_end:, patient_id_list:
+      )
+      parameters_response = FHIR::Parameters.new(
```
lmd59: Is it worth adding a new function or expanding the existing `create_parameters_response` function to allow creating a response with multiple Bundles (one for each patient)? Seems like we could use some expanded logic around creating parameters responses in a few places in these spec tests.

lmd59 (commented):
Looks to work well! Most of my comments are about rubocop style and re-use. Also, would you mind updating the PR description's test section? The testing comment says to use deqm-test-server, but I believe these tests are meant to be tested with bulk-export-server.

lib/utils/collect_data_utils.rb
ptran265: Good idea. However, when I do that, it re-enables rubocop ParameterLists check but then I have to disable Metrics/AbcSize

spec/deqm_test_kit/v1.0.0/collect_data_spec.rb
ptran265: Added in [4c613cc](https://github.com/projecttacoma/deqm-test-kit/pull/70/commits/4c613cce57bc644055af8f681a3ac8e109b8f6df)

spec/deqm_test_kit/v1.0.0/collect_data_spec.rb
```diff
@@ -263,6 +263,188 @@
...
+        parameter: patient_id_list.map do
+          FHIR::Parameters::Parameter.new(
+            name: 'return',
+            resource: FHIR::Bundle.new(
+              type: 'transaction',
+              entry: [] # no measure reports
+            )
+          )
+        end
+      )
```
lmd59:
```suggestion
      parameters_response = create_parameters_response(measure_urls: [measure_url, measure_url], patient_ids: patient_id_list)
```
Just because I think we can simplify the readability of this a bit with the new `create_parameters_response` method (creates 2 measure reports rather than none, but still valid for this test).

spec/deqm_test_kit/v1.0.0/collect_data_spec.rb
```diff
@@ -263,6 +263,188 @@
...
+                    type: 'data-collection',
+                    measure: measure_url,
+                    period: { start: period_start, end: period_end }
+                  )
+                }
+              ]
+            )
+          )
+        end
+      )
```
lmd59:
```suggestion
      parameters_response = create_parameters_response(measure_urls: [measure_url], patient_ids: patient_id_list)
```
Just because I think we can simplify the readability of this a bit with the new `create_parameters_response` method :)

spec/utils/collect_data_spec_utils.rb
```diff
@@ -5,31 +5,56 @@
 module DEQMTestKit
   # Helpers shared by $collect-data specs
   module CollectDataSpecUtils
-    def create_parameters_response(measure_urls:)
+    # rubocop:disable Metrics/MethodLength
+    def create_parameters_response(measure_urls:, patient_ids: nil)
```
lmd59: So we're not actually using the patient_ids in this method at all... we *could* use them by adding a `subject` to the measure reports, but I'm not sure there would be a huge add for testing. We're essentially just using the number of patient ids to decide the number of bundles, so I think it would be clearer to just send that number through. Something like this:

```suggestion
    def create_parameters_response(measure_urls:, bundle_count: 1)
```

lib/utils/collect_data_utils.rb
lmd59: This is a bit annoying - we could break up the function more, but I'm not sure that would improve readability. I guess I'm okay with disabling rubocop for this, but since @elsaperelli has more experience with Inferno and expected style, I might be interested in her opinion on local disables of the rubocop rules like this.

lmd59 (commented):
Thanks for the updates! Just a few more recommendations!

spec/deqm_test_kit/v1.0.0/collect_data_spec.rb
ptran265: Yes, thank you, I forgot about those fail tests!

spec/utils/collect_data_spec_utils.rb
ptran265: I applied that change to the util function, which in turn meant I needed to change the calls in the actual spec tests. I wasn't sure if you meant they should just pass that number there too or if they can use the length of the list there (so like `bundle_count: patient_id_list.length`). I ended up doing the latter because I felt that's less hardcode-y, but let me know if you disagree.

lmd59: I think that approach makes sense!

lmd59 (approved):
Looking good to me!

